PR #9, opened 2025-08-18
Implement viewer abstraction.

kevinzakka:


kevinzakka:
@brentyi I think we can still polish this further and further probably make `viser` an optional dependency if the user wants the viser viewer but I think this is fine for now and we can revisit closer to release.

PR #10, opened 2025-08-18
Add `ParamSpec` annotation to `term()` helper, `EventMode` type

brentyi:
image

----

Missing args now raise errors:

image

----

LSP completions work too:

image


kevinzakka:
Thanks Brent!

PR #20, opened 2025-08-25
Adds Unitree humanoids RL tasks

louislelay:
I’m exploring the RL and asset config parts of mj_lab to get more familiar with them. As part of this, I’m implementing locomotion tasks for H1 and G1, using assets that I know work well in both MuJoCo and Isaac Lab.

Progress might be a bit slow due to other priorities, but I wanted to open this PR as a draft so you know I’m looking into it and to share early progress/receive feedback.

### DIFF
--- a/src/mjlab/tasks/locomotion/velocity/config/g1/flat_env_cfg.py
+++ b/src/mjlab/tasks/locomotion/velocity/config/g1/flat_env_cfg.py
@@ -23,6 +23,8 @@ class UnitreeG1FlatEnvCfg(LocomotionVelocityFlatEnvCfg):
         history_length=3,
         track_air_time=True,
         filter_expr=[".*knee.*"],
-        geom_filter_expr=[".*ankle_collision"],
+        geom_filter_expr=[".*_foot3_collision"],
       ),
     }
+    self.sim.njmax = 300
+    self.sim.nconmax = 100000


kevinzakka:
Nice! Let me know if you run into any issues and how we can improve things overall. Goal is to make the transition from ISAACLab to mjlab as seamless as possible!

louislelay:
Thanks! So for the moment, I focused on the G1, since you already had the asset config file prepared for mimic. Training is currently running (I don’t know yet if it’s going well or not, since I just launched it ahah) with 4096 envs. Similarly to your mimic env, I set `njmax` to 300 and `nconmax` to 100000.

The ETA is showing 50 minutes of training, which seems very high. On Isaac Lab, I think with similar locomotion envs and the same computer I’m usually closer to 10–15 minutes. Do you have any idea why the training time is so long? If you have any pointers, I could take a look.

kevinzakka:
@louislelay Can you disable the contact sensor for now? It won't be a 1 to 1 match with Isaac but it's currently broken (see #21).

louislelay:
ok @kevinzakka!

kevinzakka:
BTW, some isaac tasks are really silly. Let’s maybe have a chat or make an issue about what tasks are worth porting over (one hard criterion being something that has actually been vetted on hardware).

louislelay:
Yes definitely worth talking abt, if you've time this week let's have a chat!

PR #24, opened 2025-08-27
Fixes video recording in training, unifies logs directory, and normalizes _HERE path

louislelay:
While encountering the following error during training with `--video`:

```
ValueError: ('Render mode is None, which is incompatible with RecordVideo.', 
'Initialize your environment with a render_mode that returns an image, such as rgb_array.')
```

I corrected it by adding the same environment creation logic as in the `play` script, with a condition on the `render_mode` based on the `video` flag.

During this change, I noticed that the `logs` directory was being created at the same level as the `train` and `play` scripts. To stay consistent with Isaac Lab, I moved it to the root of the repository.

I also updated the way `_HERE` is created (see [this explanation](https://stackoverflow.com/questions/65388213/why-is-pathlib-path-file-parent-parent-sensitive-to-my-working-directory)):

* previously it could be absolute or relative depending on how the script was called
* now it is always an absolute and normalized path using `Path(__file__).resolve().parent`

The same modifications were applied to the tracking play and train scripts, although I couldn’t test those as I don’t have a pretrained motion available

### DIFF
--- a/scripts/velocity/rl/play.py
+++ b/scripts/velocity/rl/play.py
@@ -26,8 +26,6 @@ FRAME_TIME = 1.0 / 60.0
 KEY_BACKSPACE = 259
 KEY_ENTER = 257
 
-_HERE = Path(__file__).resolve().parent
-
 
 def main(
   task: str,
@@ -52,7 +50,7 @@ def main(
   env_cfg.sim.render.height = video_height or env_cfg.sim.render.height
   env_cfg.sim.render.width = video_width or env_cfg.sim.render.width
 
-  log_root_path = _HERE.parents[2] / "logs" / "rsl_rl" / agent_cfg.experiment_name
+  log_root_path = Path("logs") / "rsl_rl" / agent_cfg.experiment_name
   log_root_path = log_root_path.resolve()
   print(f"[INFO]: Loading experiment from: {log_root_path}")
 
--- a/scripts/velocity/rl/train.py
+++ b/scripts/velocity/rl/train.py
@@ -24,8 +24,6 @@ torch.backends.cudnn.allow_tf32 = True
 torch.backends.cudnn.deterministic = False
 torch.backends.cudnn.benchmark = False
 
-_HERE = Path(__file__).resolve().parent
-
 
 def main(
   task: str,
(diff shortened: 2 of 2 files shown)


louislelay:
Hey @kevinzakka, I’m running into a strange loop with the Ruff checks!

Locally everything looks clean after running Ruff + pre-commit:

```
$ uvx ruff@0.12.10 check --fix .
All checks passed!

$ pre-commit run --all-files
ruff check...............................................................Passed
ruff format..............................................................Passed
```

But on the PR checks I consistently see errors like:

```
--- src/mjlab/tasks/tracking/rl/runner.py
+++ src/mjlab/tasks/tracking/rl/runner.py
@@ -1,9 +1,9 @@
 import os

+import wandb
 from rsl_rl.env.vec_env import VecEnv
 from rsl_rl.runners import OnPolicyRunner

-import wandb
 from mjlab.rl import RslRlVecEnvWrapper
...
Would fix 2 errors.
Error: Process completed with exit code 1.
```

So CI wants one import ordering, while pre-commit locally forces the opposite. When I try to commit what CI seems to want, the pre-commit hook refuses the commit and rewrites the file back:

```
$ git commit -m "formats to pass PR checks"
ruff check...............................................................Failed
- hook id: ruff-check
- files were modified by this hook
```

After staging those changes and committing again, locally everything passes, but CI then fails again with the other version.

Not sure if this is a config/version mismatch between pre-commit and CI, or something wrong in my PR itself. Maybe you already know what’s going on here and can point me in the right direction so I can get the PR checks to pass.

kevinzakka:
@louislelay Delete the wandb directory in the src directory and it will fix the ruff issue :)

louislelay:
Thank you @kevinzakka! :laughing:

scripts/tracking/rl/play.py
```diff
@@ -45,7 +45,7 @@ def main(
   env_cfg.sim.render.height = video_height or env_cfg.sim.render.height
   env_cfg.sim.render.width = video_width or env_cfg.sim.render.width
 
-  log_root_path = _HERE / "logs" / "rsl_rl" / agent_cfg.experiment_name
+  log_root_path = _HERE.parents[2] / "logs" / "rsl_rl" / agent_cfg.experiment_name
```
kevinzakka: I think you should be able to just do Path("logs") / "rsl_rl" / ... for log_root_path. Can you confirm?

louislelay: Yes, that's right! I corrected it in https://github.com/mujocolab/mjlab/pull/33 (I lost write access so I can't push on this branch anymore)

PR #27, opened 2025-08-29
Add Pyright

brentyi:
I added Pyright and started fixing type errors.

On `main`:
```
~/mjlab uv run pyright
[...]
434 errors, 0 warnings, 0 informations
```

On this branch:
```
~/mjlab uv run pyright
0 errors, 0 warnings, 0 informations
```

It seems like most of these were real issues. I tried my best to fix, but there were a few things I wasn't sure about:
- The correct type for indexing when resetting. We sometimes had `torch.Tensor | slice | None` in the type signature, but didn't actually support all of them at runtime. Should we standardize?
- The `.func` attribute. This was generally annotated as `func: Callable`, but we sometimes also call `.func.reset(env_ids)`.
- Incompatible overrides of `gym.Env`, for example returning tensors instead of ints. I see why this makes sense, I guess the only solutions is `type: ignore`?


kevinzakka:
Nice, pyright passes for me too now!

PR #28, opened 2025-08-29
Bugfix for domain randomization.

kevinzakka:
We were previously only expanding fields for domain randomization events that happen at startup.


src/mjlab/sim/sim.py
qiayuanl, line 1: Maybe move the table printing to the `event_manager.py`

kevinzakka: Done, thanks!

PR #29, opened 2025-09-01
Add reward plots to native mujoco visualizer.

kevinzakka:
This PR does 2 things:

1. Cleaned up the base viewer class, removing some native-specific methods. This allows us to remove them from the `viser` implementation.
2. Added a small info text overlay showing the current env index, episode step, playback speed and FPS.
3. Added a live visualization of the rewards which can be toggled with the `P` key. See the screenshot below.

Screenshot From 2025-09-01 15-26-22

### DIFF
--- a/src/mjlab/viewer/base.py
+++ b/src/mjlab/viewer/base.py
@@ -75,7 +75,7 @@ class ViewerAction(Enum):
 
 
 class BaseViewer(ABC):
-  """Abstract base class for environment viewers with a minimal shared API."""
+  """Abstract base class for environment viewers."""
 
   SPEED_MULTIPLIERS = [0.01, 0.016, 0.025, 0.04, 0.063, 0.1, 0.16, 0.25, 0.4, 0.63, 1.0]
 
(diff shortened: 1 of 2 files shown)


kevinzakka:
cc @brentyi

yuvaltassa:
Super nice!!
Consider one click up on the default [mjtFontScale](https://mujoco.readthedocs.io/en/latest/APIreference/APItypes.html#mjtfontscale)

kevinzakka:
@yuvaltassa Weirdly, no matter what value I set, I get the same font size:

```python
overlay = (
  mujoco.mjtFontScale.mjFONTSCALE_150.value,  # tried 150, 200, 300.
  mujoco.mjtGridPos.mjGRID_TOPLEFT.value,
  text_1,
  text_2,
)
viewer.set_texts(overlay)
```

yuvaltassa:
I think the font size is determined once when you [make the GL context](https://mujoco.readthedocs.io/en/latest/APIreference/APIfunctions.html#mjr-makecontext), though you can also change it [post hoc](https://mujoco.readthedocs.io/en/latest/APIreference/APIfunctions.html#mjr-changefont)

PR #37, opened 2025-09-06
Fix viewer memory leak; more features for Viser viewer

brentyi:
Visualization-related changes. Work-in-progress, but it should run and the changes are relatively self-contained.

Command I'm using:
```
uv run --extra rl scripts/tracking/rl/play.py \
  --task Mjlab-Tracking-Flat-G1-Play-v0 \
  --wandb-run-path gcbc_researchers/mjlab_alpha/rfdej55h \
  --viewer viser \
  --num-envs 1024
```

Changes:

**(1)** Memory leak. The base viewer steps was leaking GPU memory for me during every environment step. This was causing crashes if I left the viewer running for too long. I didn't analyze this in-depth, but wrapping everything in `torch.inference_mode()` fixed this for me. Is this expected @kevinzakka?

**(2)** General Viser GUI improvements. We now have a basic info panel, simulation controls, slider for environment spacing:

https://github.com/user-attachments/assets/cb9c5cc5-c2e1-4809-99b2-1c77d83207dd

I *think* the jitter here is from the discretization in the env spacing slider that I'm dragging, and not actually the updates/rendering.

**(3)** The visual geoms also now support texture. This is internally done by converting the Mujoco format => trimesh => GLB, which Viser can visualize:

https://github.com/user-attachments/assets/55a8f12f-1c79-43f1-a7a6-1d0bb9e80978

**(4)** I also added some proof-of-concept reward plots in, inspired by #29. There's a slider for selecting which environment to visualize rewards for:

https://github.com/user-attachments/assets/a4460c17-9453-4727-87c1-b3c5e6ae9411

**Possible issue:** I was sometimes seeing framerates drop when "Show only this environment" is checked... this is counterintuitive. I can investigate and (probably) fix in the next Viser release.

### DIFF
--- a/src/mjlab/viewer/viser.py
+++ b/src/mjlab/viewer/viser.py
@@ -80,7 +80,7 @@ class ViserViewer(BaseViewer):
         self._status_html = self._server.gui.add_html("")
 
       # Simulation controls
-      with self._server.gui.add_folder("Simulation Controls"):
+      with self._server.gui.add_folder("Simulation"):
         # Play/Pause button
         self._pause_button = self._server.gui.add_button(
           "Play" if self._is_paused else "Pause",


src/mjlab/viewer/base.py
```diff
@@ -171,12 +171,14 @@ def request_action(self, name: str, payload: Optional[Any] = None) -> None:
   def step_simulation(self) -> None:
     if self._is_paused:
       return
-    with self._sim_timer.measure_time():
-      obs = self.env.get_observations()
-      actions = self.policy(obs)
-      self.env.step(actions)
-      self._step_count += 1
-    self._accumulated_sim_time += self._sim_timer.measured_time
+    # Wrap in inference_mode to prevent gradient accumulation and memory leaks
```
kevinzakka, line 174: This makes sense, it's present in the isaac play script and I forgot to port it over here. Nice catch!

scripts/tracking/rl/play.py
```diff
@@ -83,8 +85,12 @@ def main(
 
   policy = runner.get_inference_policy(device=env.device)
 
-  viewer = NativeMujocoViewer(env, policy, render_all_envs=render_all_envs)
-  viewer.run()
+  if viewer == "native":
```
kevinzakka, line 88: Thanks for adding this!

src/mjlab/viewer/viser.py
```diff
@@ -84,32 +255,61 @@ def _(_) -> None:
...
 
-    # Skip every other frame to reduce load. 30FPS is fine!
+    # Update counter
     self._counter += 1
+
+    # Update status display less frequently (every 10 frames = ~6 updates per second)
+    if self._counter % 10 == 0:
+      self._update_status_display()
+
+    # Skip every other frame for position updates to reduce load. 30FPS is fine!
```
kevinzakka: Do you want to just make the default FPS 30? On the native viewer, 30 feels choppy which is why I opted for 60.

brentyi: Actually, updates at 60 FPS should work fine. I'll just remove this.

PR #44, opened 2025-09-16
Command-line overrides for agent and env configs

brentyi:
Changes:
- Updated the training scripts to allow command-line overrides for agent and env configs.
- Task is now a positional argument (no more `--task`).
     - We could also keep it backwards-compatible if you prefer @kevinzakka.
- Removed args in train scripts that can be set by overriding the agent/env config directly. (example: `--num-envs` is now `--env.scene.num-envs`)
- Removed ` device` fields from configs. This is mostly an aesthetic thing, since it forces you to do a lot of config mutation.

Questions:
- Should we make a similar change to `play.py`? Alternatively, can `play.py` just load the config / task from wandb?

Helptext when directly running `train.py`:

image

From `scripts/tracking/rl/train.py Mjlab-Tracking-Flat-G1-v0 --help`:

image

[ for better or worse this is followed by a few more pages of helptext... ]

### Additional notes

Training commands have changed slightly.

*Before:*
```bash
MUJOCO_GL=egl uv run --extra rl scripts/tracking/rl/train.py \
  --task Mjlab-Tracking-Flat-G1-v0 \
  --registry-name gcbc_researchers/csv_to_npz/lafan_cartwheel \
  --num_envs 4096
```

*After:*
```bash
MUJOCO_GL=egl uv run --extra rl scripts/tracking/rl/train.py \
  Mjlab-Tracking-Flat-G1-v0 \
  --registry-name gcbc_researchers/csv_to_npz/lafan_cartwheel \
  --env.scene.num-envs 4096
```

The helptext can also take many seconds to appear after passing in `--help`. This is because all of the imports need to happen before `tyro` runs, and these are slow:

image

Lazy imports might be a good idea in the future.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,6 +99,7 @@ mujoco = { index = "mujoco" }
 torch = [{ index = "pytorch-cu128", extra = "cu12" }]
 
 [tool.ruff]
+src = ["src"]  # Helpful for recognizing first-party imports.
 indent-width = 2
 exclude = [
   "src/mjlab/third_party",
--- a/scripts/tracking/rl/play.py
+++ b/scripts/tracking/rl/play.py
@@ -4,9 +4,9 @@ from typing import Literal, cast
 
 import gymnasium as gym
 import tyro
+import wandb
 from typing_extensions import assert_never
 
-import wandb
 from mjlab.rl import RslRlOnPolicyRunnerCfg, RslRlVecEnvWrapper
 from mjlab.tasks.tracking.rl import MotionTrackingOnPolicyRunner
 from mjlab.tasks.tracking.tracking_env_cfg import TrackingEnvCfg


kevinzakka:
No need to make it backwards-compatible!

src/mjlab/managers/action_manager.py
```diff
@@ -136,6 +136,8 @@ def _prepare_terms(self):
     self._term_names: list[str] = list()
     self._terms: dict[str, ActionTerm] = dict()
 
+    from mjlab.managers.manager_term_config import ActionTermCfg
```
kevinzakka, line 139: Why put it in type checking and also as a local import?

brentyi: Avoiding a circular import 🙃 

- We previously had `type[ActionTerm]` with the `ActionTerm` in `TYPE_CHECKING` guard in the config source, but this breaks with `tyro` because `tyro` needs to be able to retrieve the type annotations at runtime
- After removing the `TYPE_CHECKING` guard in the config code, this file now needs the guard to prevent the circular import (it's used as an annotation above)
- This method needs `ActionTermCfg` at runtime for `get_terms()`, so we need to import it for real here

I think this double-import will go away when we revisit `get_terms()`!

kevinzakka: Cool! Thank you!

PR #60, opened 2025-09-20
Adds first version of a helper script

louislelay:
This PR adds a first version of a helper script (`mjlab.sh`) to make environment setup and management easier, inspired by how Isaac Lab handles its own envs. It installs `uv` if missing, creates a dedicated `env_mjlab` virtual environment, and sets up handy aliases (`env_mjlab` to activate, `mjlab` to run project scripts). The idea is to keep things simple for users while making sure everyone runs inside a consistent virtual environment, which also helps later with external project generator work.

Down the line, the script could grow flags for running tests, formatting, generating external projects, or even building docs (e.g. with Sphinx). For now it’s Linux-only and intentionally minimal, I just wanted to get a working first iteration and see you'd find it interesting. Extra flags should be easy to add, but Windows support will take longer (I don’t even have a Windows machine to try it on ahah).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,3 @@ motions/
 artifacts/
 .venv/
 env_mjlab
-uv.lock
--- a/uv.lock
+++ b/uv.lock
@@ -1,0 +1,3144 @@
+version = 1
+revision = 3
+requires-python = ">=3.10"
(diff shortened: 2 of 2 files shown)


louislelay:
Closing it after internal discussions

PR #64, opened 2025-09-21
Add null command and curriculum managers.

kevinzakka:
@louislelay I think this might be cleaner. Let me know what you think.


louislelay:
LGTM!

Issue #61, opened 2025-09-20
Add an external project generator


louislelay:
Add a way to make a new project with basic files. This lets users build tasks outside the main repo, in a clean, self-contained way.

PR #69, opened 2025-09-21
Adds external project generator

louislelay:
Implementation of an external project generator with built-in packages. To use it:

```bash
python scripts/generator.py
```

It will create a new project and copy into it a Go1 task example along with its asset. Then you're good to go:

```
cd path/to/ext_project
uv run scripts/list_envs.py
```

Closes https://github.com/mujocolab/mjlab/issues/61

Future work:
- Align naming conventions: task IDs (Project-Name-...), package name (project_name), and constant (PROJECT_NAME_SRC_PATH). Currently, all use the package name directly.
- Add option to choose the path for the external project
- Support choice between manager-based and direct environments (once direct envs are implemented ofc)

I think all this can be done after merging, but def open to opinions!


louislelay:
I need to updates some paths due to recent changes

louislelay:
Let’s pause this until https://github.com/mujocolab/mjlab/pull/84 is reviewed/merged.

PR #124, opened 2025-10-01
Fewer classes in config organization (again)

brentyi:
Ways we currently use classes in configs include:
- As key-value stores. We define dataclasses where each field is some kind of action/observation/reward/etc term, and then call a `get_terms()` helper to list out terms.
- For config inheritance. A Go1 velocity config inherits some fields from a base velocity config, which inherits some fields from the base ManagerBasedRlEnvConfig.
    - Some overrides are done by overriding dataclass fields. This feels slightly brittle but fine.
    - Other overrides are done in `__post_init__()`. This is problematic since it overwrites any inputs passed to the class constructor. (eg by tyro)

In this PR, we:
- Switch to dictionaries whenever key-value pairs are needed. 
- Flatten the config hierarchy. All configs are now `ManagerBasedRlEnvConfig`. 

I think this is more reasonable / ergonomic than needing to define lots of new classes for each new task, but it will be less familiar for folks who are used to IsaacLab. Interested in thoughts!

(revival of #97)


louislelay (commented):
Hey @brentyi, I get the motivation and like the simplification, but I’m a bit mixed. Even if the changes aren’t huge, they chip away at the “just drop in tasks from Isaac Lab” feel, before you could almost copy/paste. Curious what @kevinzakka thinks. In any case, if it’s merged, you should update the migration guide! :)

brentyi:
Totally reasonable. My personal feeling is that it's worth the change and migration is still easy, especially with Cursor/Claude/etc, but this change is fairly subjective. (to disclose my bias I also don't personally care much about adhering to the Isaac API)

If we decide against merging this I can make an alternate PR that just removes all of the `__post_init__()` stuff, since that currently prevents options from being overridden during config construction and from the command-line.

brentyi:
@kevinzakka FYI just updated the PR description with more explanation

kevinzakka:
I'm leaning towards approving this. Writing correct code is important and we don't have to inherit Isaac's "broken" mechanism. The only downside is that it would make it more difficult for users to port their code over.

@qiayuanl What are your thoughts?

romesco:
Thank you for caring about this! I love IsaacLab and all of its thoughtful design decisions from over the years ( big up to the team ! ), but the right UX is always an ongoing experiment.

I've put a lot of thought into this too, and what I've realized is there are two concerns that can be dealt with separately, but are often pitted against each other in the same part of a codebase, effectively in a perpetual tug-of-war. These are:

A. Code Structure and Validation. IMO, this should have the properties of:
> 1. Correctness (as the highest concern) - e.g. what you ask for is what you get (no `__post_init__` surprise)
> 2. Guardrails / Validation - Limits the configurable space to what makes sense (e.g. as much strong type-checking as possible)
> 3. Standard Python - Minimize custom infrastructure code (e.g. remove specialized `configclass`, `term()`, etc.). A repo should not have to maintain special meta-datastructures

B. User Experience. With the properties:
> 1. Tersely summarizes what is available to play with.
> 2. Discoverability for digging deeper past what is surfaced by default.
> 3. Flexible to different UIs: e.g. CLI, yaml overrides, GUI sliders, etc.)

In general, I believe it is more important to build the codebase around (A) and provide a lens onto it which is distinct and satisfies (B). A lens can be a "higher level description" that compiles to the layer that is explicit and prioritizes (A). This is especially relevant now that on-the-fly code generation is becoming feasible, and the main concern there is that the generated code is validated. IMO, separating a codebase in this way is important, otherwise there will alway be a (more A)->(more B)->(more A) tug-of-war.

@brentyi  I am curious what your experience with respect to this has been for nerfstudio - I haven't tracked the opensource life of the repo nearly as closely as Isaac/MuJoCo

---

In an effort to be a little more concrete and respond directly to the PR:

I (as we well as collaborators and mentees)  have often encountered the `__post_init__` resolution confusion as well as the "mispell -> silent error" issue. It has always just been something you have to "cope with". It would be great to remove these failure modes.

Towards (A), these changes make a lot of sense and I support them.  @louislelay 's comment about  losing the drop-in ability is also definitely valid. I think there's a way to both improve the separation of (A) and (B) where the drop-in aspect falls under (B). It would be nice at some point to have a way to "compile" an IsaacLab task representation to look like what ends up best satisfying (A).

~~My only other comment is that it might be good to consider `TypedDict` instead of plain dict. This will offer more in the way of static type checking with minimal extra work.~~ I thought about this more, and it seems like it will still have the drawbacks of `dataclass` when adding terms. So nevermind that suggestion 🙂

brentyi:
@romesco really appreciate the note and agree with everything you wrote!

My feeling for projects like `mjlab` is that code maintainability and (A) should be optimized above basically all else. To me this isn't actually in conflict with user experience and (B), although it's often in conflict with familiarity and common practices for achieving (B).

> I am curious what your experience with respect to this has been for nerfstudio - I haven't tracked the opensource life of the repo nearly as closely as Isaac/MuJoCo

As a disclaimer on NS I didn't design much of the codebase, but I do think we got a lot of things right. Wrt configs, NS was originally built around a library of YAML files... switching to configs that can be type-checked with the rest of the codebase was way less error-prone and made maintenance, merging changes, etc much easier.

Things that eventually made life difficult: trying to support so many things in the main repo, designing NS as a library without being very clear about version pinning + backwards compatibility expectations, etc. Probably there are some lessons that transfer to `mjlab`. 

> It would be nice at some point to have a way to "compile" an IsaacLab task representation to look like what ends up best satisfying (A).

I do think this kind of thing is now much easier with Cursor/Claude/etc! (especially with tool-calling for inspecting types, running pyright, etc, but consistent success possibly requires some prompting skill)

---

On this PR: from discussions with @kevinzakka it sounds like the plan is to merge. I'll try to resolve the conflicts, see if there's any docs that need updating, etc, in the coming few days.

jaraujo98:
@brentyi I've been looking forward to seeing this PR merged. In the meantime, I’d like to offer one suggestion.

One advantage of the current configuration class interface is that we can access everything using dot notation. By switching to dictionaries, we’ll need to use string indexing instead. For example:

```python
cfg.observations.policy.enable_corruption = False
```

would become

```python
cfg.observations["policy"].enable_corruption = False
```

This not only introduces a lot of code changes but also makes the syntax a bit more cumbersome (`"['...']"` vs ".").

Would it be possible to preserve the existing interface by using a custom dictionary class, such as:

```python
class configdict(dict):
    def __getattr__(self, attr):
        return self[attr]
    def __setattr__(self, attr, value):
        self[attr] = value
```

Do you see any drawbacks with this approach? Would it conflict with the design principles motivating the switch to plain dictionaries?

brentyi:
> Do you see any drawbacks with this approach? Would it conflict with the design principles motivating the switch to plain dictionaries?

Aesthetically I'm also mixed on the string indexing. Functionally though, I prefer it because it reduces indirection and produces a more transparent API (code should be honest, etc). For my own workflow, with
```python
cfg.observations["policy"].enable_corruption = False
```
, I appreciate being able to tell immediately what's a true attribute (supports jump-to-def, will autocomplete, be type-checked, etc) and what's a dynamic string (where the programmer has a higher burden to avoid typos, etc).

Related also is this goal from @romesco:
> Minimize custom infrastructure code (e.g. remove specialized `configclass`, `term()`, etc.). A repo should not have to maintain special meta-datastructures

That said, I know this kind of syntactic sugar is common and people like it in libraries like Hydra and ml_collections. It's possible it would be more familiar to people?

rdeits-bd:
Just in case this perspective is useful, I'm in the middle of porting a bunch of IsaacLab code to mjlab right now. I basically fundamentally agree with the motivation for this PR -- I've found the dataclass-as-dict pattern to be unwieldy both in IsaacLab and here, and I've had several bugs already due to accidentally forgetting to add a type annotation to a field in my `RewardsCfg`, which silently causes that reward to be ignored because it doesn't become a proper dataclass `field`. Using dicts would make that kind of bug impossible. 

Another thing I'm finding awkward with the dataclass API is configuring tasks (e.g. for two different flavors of robot) functionally rather than through mutation. If, for example, one reward requires knowing the names of the foot links of my robot, I'd love to be able to pass in that data where I construct the reward, rather than adding a `post_init` mutation step several layers of class hierarchy up. I could maybe sort of use `InitVar` for this, but it's pretty painful and still requires mutation in `post_init`. 

I have no idea if this will make it easier or harder to "sell" mjlab as an isaaclab replacement -- for me I think it would be no particular burden, as the actual number of classes I'm making is really quite small. I do have a slight preference towards @jaraujo98's suggestion to wrap the dict in a convenience accessor so that the dot notation just works, but ultimately it's up to y'all and I'll be able to work with whichever API you pick.

brentyi:
Thanks @rdeits-bd and everyone else for the helpful comments! Going to close this PR in favor of #292. @kevinzakka hopefully we can merge in the next day or so 🤞

PR #127, opened 2025-10-01
add exporting logic to play

WT-MM:
Matches IsaacLab convention by adding a flag to export the policy during `play`.


kevinzakka:
Hi @WT-MM, do we really need this if exporting already happens during training?

kevinzakka:
Going to close this PR since I'm cleaning up clutter. Feel free to re-open if you feel strongly about this.

PR #132, opened 2025-10-02
Update runner.py, fix path handling in runner.py for Windows compatibility

Hao-yang-Wang:
The code previously used `split("/")` which breaks on Windows  because of backslashes in file paths. 
This patch uses `os.path` for cross-platform compatibility.


kevinzakka:
Thanks @Hao-yang-Wang!

PR #138, opened 2025-10-03
Refactor name resolution methods in SceneEntityCfg

CharlieLeee:
also add clear error message.


kevinzakka:
Hi @CharlieLeee, I would be comfortable merging this if we have unit tests dedicated to `SceneEntityCfg`. I'll try to write some over the weekend and then we can make sure your refactor does not break anything!

PR #139, opened 2025-10-03
Update play.py

Hao-yang-Wang:
Fix motion_file requirement logic in run_play. 

**Changes**
Only require motion_file when running Tracking tasks (TrackingEnvCfg). Non-Tracking tasks can now load checkpoints without needing motion_file.

**Why**
Previously, motion_file was always required with a checkpoint, which caused issues for non-Tracking tasks.

### DIFF
--- a/src/mjlab/scripts/play.py
+++ b/src/mjlab/scripts/play.py
@@ -43,15 +43,18 @@ def run_play(
     device = "cuda:0" if torch.cuda.is_available() else "cpu"
   print(f"[INFO]: Using device: {device}")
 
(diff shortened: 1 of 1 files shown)


kevinzakka:
Right I had this fix in a different branch but never merged. Thanks @Hao-yang-Wang! Mind running `make format` or setting up the pre-commit hook?

Hao-yang-Wang:
Sure! I’ll run make format and update the PR.

kevinzakka (approved):

Issue #122, opened 2025-09-30
Unify the arguments between play and train scripts


louislelay:
It would be nicer to have the same arguments for play than for train.

PR #158, opened 2025-10-08
Unifies arguments between play and train scripts

louislelay:
This modifies the play script to handle the argument the same way as the train script (removing the need for `--task`). I also added `list_envs` as a project script, so we can now run `uv run list_envs`.

Closes https://github.com/mujocolab/mjlab/issues/122

### DIFF
--- a/src/mjlab/scripts/demo.py
+++ b/src/mjlab/scripts/demo.py
@@ -4,12 +4,8 @@ This demo downloads a pretrained checkpoint and motion file from cloud storage
 and launches an interactive viewer with a humanoid robot performing a cartwheel.
 """
 
-from functools import partial
-
-import tyro
-
 from mjlab.scripts.gcs import ensure_default_checkpoint, ensure_default_motion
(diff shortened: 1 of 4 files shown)


src/mjlab/scripts/demo.py
```diff
@@ -24,18 +20,16 @@ def main() -> None:
...
-    partial(
-      run_play,
-      task="Mjlab-Tracking-Flat-Unitree-G1-Play",
-      checkpoint_file=checkpoint_path,
-      motion_file=motion_path,
-      num_envs=8,
-      render_all_envs=True,
-      viewer="viser",
-    )
+  play_config = PlayConfig(
```
brentyi, line 25: perhaps we can do `tyro.cli(PlayConfig, default=play_config)`?

That way we can keep the `--num-envs`, `--viewer`, etc available as arguments

louislelay: nice idea! just tested and it works well

PR #183, opened 2025-10-14
NaN Guard for Debugging Numerical Instability

### LINKED ISSUE #46: Add NaN guard with rolling state buffer and conditional dump on NaN/Inf
During training we occasionally hit `NaN`s, but we can’t provide reproducible state to the mjwarp team. Add a lightweight “NaN Guard” that (1) keeps a rolling buffer of the last K training states and (2) writes them to disk when `NaN`/`Inf` is detected in `wp.Data`. The guard must be disabled by default and cheap when off.

kevinzakka:
Adds a lightweight NaN guard system that captures simulation states in a rolling buffer and dumps them to disk when NaN/Inf is detected. Includes an interactive Viser-based viewer to scrub through captured states.

https://github.com/user-attachments/assets/81818528-47b6-4171-bba3-4229df1433a1

*Interactive NaN dump viewer: scrub through captured states to debug numerical instability. Env 1 shows broken geometry (NaN), env 2 is clean, env 3 teleported away (Inf).*

### Features

- Automatic capture with rolling buffer (zero overhead when disabled)
- Saves state history (npz) and model (mjb) when NaN/Inf detected
- Interactive viewer (`viz-nan`) to scrub through states with sliders
- Train integration via `--enable-nan-guard` flag

### Usage

Enable during training:
```bash
uv run train Mjlab-Velocity-Rough-Unitree-Go1-v0 --enable-nan-guard
```

View captured states:
```bash
uv run viz-nan /tmp/mjlab/nan_dumps/nan_dump_20251014_095857.npz
```

The viewer shows which environments crashed and lets you compare them against clean environments in the same capture.

Fixes #46.


src/mjlab/scripts/nan_viz.py
```diff
@@ -0,0 +1,250 @@
...
+      </div>
+    """
+
+  def _setup_scene(self) -> None:
+    """Setup the 3D scene with the model."""
+    self.server.scene.configure_environment_map(environment_intensity=0.8)
+    self._add_fixed_geometry()
+    self._create_body_meshes()
+
+  def _add_fixed_geometry(self) -> None:
```
brentyi, line 124: this is currently repeated between the main viewer and NaN vis tool, right? should we consolidate somehow?

kevinzakka: Consolidated some shared functions between both viewers.

brentyi (approved):

PR #200, opened 2025-10-18
Add effort control mode to actuator cfg

martinorrje:
Adds the option to specify control mode as either "effort" or "position" in the actuator cfg. Default is set to position, ensuring compatibility with already created asset configs.


kevinzakka:
Hi @martinorrje, I'm planning a re-factor of actuators in the coming 2 weeks and will address this. Thanks!

PR #207, opened 2025-10-20
Fixes init keyframe ctrl sizing, aligns joint-action indexing and adds tests

louislelay:
This PR fixes a crash that happened when creating the `init_state` keyframe for articulated entities. Previously we were doing:

```python
key.ctrl = joint_pos
```

which assumes the number of actuators matches the joints. When actuator configs changed, it failed with:

```
ValueError: Error: Keyframe init_state has invalid ctrl size, got 26, should be 22
```

I now build `ctrl` with the correct width and map values per actuator:

```python
ctrl = np.zeros(len(self._spec.actuators))
for i, act in enumerate(self._spec.actuators):
    try:
        jidx = self.joint_names.index(act.name)
        ctrl[i] = joint_pos[jidx]
    except ValueError:
        ctrl[i] = 0.0
key.ctrl = ctrl
```

This removes the 1:1 joint↔actuator assumption and avoids the size mismatch. I also fixed `joint_actions` to index offsets by `joint_ids` instead of `actuator_ids`.

### DIFF
--- a/src/mjlab/entity/entity.py
+++ b/src/mjlab/entity/entity.py
@@ -158,13 +158,8 @@ class Entity:
     key = self._spec.add_key(name="init_state", qpos=key_qpos)
 
     if self.is_actuated and joint_pos is not None:
-      ctrl = np.zeros(len(self._spec.actuators))
-      for i, act in enumerate(self._spec.actuators):
-        try:
-          jidx = self.joint_names.index(act.name)
-          ctrl[i] = joint_pos[jidx]
-        except ValueError:
-          ctrl[i] = 0.0
+      name_to_pos = {name: joint_pos[i] for i, name in enumerate(self.joint_names)}
+      ctrl = np.array([name_to_pos.get(act.name, 0.0) for act in self._spec.actuators])
       key.ctrl = ctrl
 
   # Attributes.
--- a/tests/test_entity.py
+++ b/tests/test_entity.py
@@ -435,10 +435,10 @@ class TestInitStateKeyframe:
     """key.ctrl length == nu and values follow actuator->joint mapping by name."""
     ent = articulated_entity_with_init
 
-    # Compile to check MuJoCo accepts the keyframe and expose key_ctrl
+    # Compile to check MuJoCo accepts the keyframe and expose key_ctrl.
     mj_model = ent.compile()
 
-    # Sanity: exactly one keyframe (the 'init_state' we add)
+    # Sanity: exactly one keyframe (the 'init_state' we add).
     assert mj_model.nkey == 1
 
     # key_ctrl is (nkey, nu)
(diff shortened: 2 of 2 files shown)


src/mjlab/entity/entity.py
```diff
@@ -158,7 +158,14 @@ def _add_initial_state_keyframe(self) -> None:
     key = self._spec.add_key(name="init_state", qpos=key_qpos)
 
     if self.is_actuated and joint_pos is not None:
```
kevinzakka, line 160: Can we write this more simply as:

```python
if self.is_actuated and joint_pos is not None:
    name_to_pos = {name: joint_pos[i] for i, name in enumerate(self.joint_names)}
    ctrl = np.array([name_to_pos.get(act.name, 0.0) for act in self._spec.actuators])
    key.ctrl = ctrl
```

tests/test_entity.py
```diff
@@ -388,5 +389,90 @@ def test_large_force_stability(self, initialized_floating_entity):
     assert not torch.any(torch.isnan(sim.data.qpos)), "Should not produce NaN"
 
 
+class TestInitStateKeyframe:
```
kevinzakka, line 392: Thanks for the tests! One minor nit: Can you add a period to comments?

tests/test_entity.py
```diff
@@ -388,5 +389,90 @@ def test_large_force_stability(self, initialized_floating_entity):
...
+
+    # Build expected ctrl
+    joint_pos_vec = np.array([0.5, -0.25])  # ["joint1", "joint2"] order
+    expected = np.zeros(mj_model.nu)
+    for i, act in enumerate(ent.spec.actuators):
+      jidx = ent.joint_names.index(act.name)
+      expected[i] = joint_pos_vec[jidx]
+
+    # Compare against compiled model keyframe ctrl row 0
+    np.testing.assert_allclose(mj_model.key_ctrl[0], expected, atol=1e-8)
```
kevinzakka: Alternatively, since we know it gets added as "init_state", you can do `mj_model.key("init_state").ctrl`

tests/test_entity.py
```diff
@@ -388,5 +389,90 @@ def test_large_force_stability(self, initialized_floating_entity):
...
+
+    # nu == number of actuators (1) and ctrl has matching width
+    assert mj_model.nu == 1
+    assert mj_model.key_ctrl.shape == (1, 1)
+
+    expected = np.array([0.42], dtype=float)
+    np.testing.assert_allclose(mj_model.key_ctrl[0], expected, atol=1e-8)
+
+  def test_keyframe_ctrl_nonzero_does_not_raise(self, articulated_entity_with_init):
+    ent = articulated_entity_with_init
```
kevinzakka: Maybe `articulated_entity_with_init.compile()`?

louislelay: done!

tests/test_entity.py
louislelay: done!

tests/test_entity.py
louislelay: done!

src/mjlab/entity/entity.py
louislelay: i like it very much, done!

PR #206, opened 2025-10-20
Sensor API Redesign

kevinzakka:
This refactor makes sensors first-class, scene-level constructs that can reference multiple entities and return typed, structured data. It lays the groundwork for upcoming RGB/Depth camera sensors.

## Changes

- Added a base Sensor with lifecycle hooks: `edit_spec`, `initialize`, `reset`, `update`.
- Added two implementations:
  - `BuiltinSensor`: wraps MuJoCo's native sensor types (gyro, accelerometer, etc.) and exposes raw sensordata.
  - `ContactSensor`: detects contacts between bodies, geoms, or subtrees. Use regex patterns to match multiple objects at once (e.g., all four feet of a quadruped), and get back clean tensors with forces, contact flags, air time related quantities, etc.

## Configuration

Sensors are declared alongside entities in the scene config. Each `*SensorCfg` is a dataclass; the engine instantiates sensors during scene building and wires `sensor.data.*` at runtime.

```python
from dataclasses import replace

scene.entities = {"robot": replace(G1_ROBOT_CFG)}

self_collision_cfg = ContactSensorCfg(
    name="self_collision",
    primary=ContactMatch(mode="subtree", pattern="pelvis", entity="robot"),
    secondary=ContactMatch(mode="subtree", pattern="pelvis", entity="robot"),
    fields=("found",),
    reduce="none",
    num_slots=1,
)
scene.sensors = (self_collision_cfg,)
```

At runtime, access the sensor by name:

```python
self_collision: ContactSensor = env.scene["self_collision"]
# With num_slots=1, shape is [B, 1], squeeze to get [B]
in_contact = self_collision.data.found.squeeze(-1)
```

Notes:
- `scene.sensors` accepts a tuple/list of sensor configs; patterns are expanded at build time into one MuJoCo sensor per (match, field) pair, then aggregated.
- The name becomes the lookup key (`env.scene["<name>"]`).

### Understanding num_slots

The `num_slots` parameter controls how many contacts to report per primary match. This is useful when a single geom/body may have multiple simultaneous contact points.

**Common use cases**:

- `num_slots=1` (most common): Track whether contact exists and get a single representative contact point
  - Best for simple contact detection (is foot touching ground?)
  - Fastest option
  - Use with reduce="maxforce" to get the strongest contact or reduce="mindist" for deepest penetration
- `num_slots > 1`: Track multiple contact points per geom/body
  - Useful when you need detailed contact information (e.g., detect multiple contacts on a body)
[...]

PR #209, opened 2025-10-20
Fixing cartpole example

Gregwar:
Hello,

This PR addresses the following  issues in [create_new_task.md](https://github.com/mujocolab/mjlab/blob/main/docs/create_new_task.md):

* Fixes inconsistent indentation
* Fixes the `play` command
* More importantly, the current tutorial is misleading on the way actuators are handled. An actuator is present in the proposed `.xml` file, and an additional one is added programmatically through mjlab with `stiffness=0.0`. From what I understand, this extra actuator has no effect
  * However, the `init_state` keyframe, automatically added by `_add_initial_state_keyframe`, sets a `ctrl` with the same size as `joint_pos` (because of `key.ctrl = joint_pos`). I believe this should not be the case since `len(ctrl) != len(joint_pos)` in general
  * Thanks to the useless extra actuator, the proposed tutorial works because `ctrl` become of size `2` like `joint_pos`, but I don't think this is a proper solution
  * In this PR, a slight workaround is added such that `init_state` is not created if it exists, I believe there is either a misunderstanding on my side, or that a deeper fix/configuration need to be added, please let me know
  * In this PR, I let the actuator in the `.xml` file and removed the definition in the code (along with the note about it that become irrelevant)

Some suggestions

* Maybe `InitialStateCfg` should contain a `ctrl` field, allowing to specify the `ctrl` value for the initial keyframe manually ?
* It could also be useful to have an option to name a keyframe from the `xml` to be used instead of the `InitialStateCfg` contents (or alternatively a field in `InitialStateCfg` to specify

### DIFF
--- a/docs/create_new_task.md
+++ b/docs/create_new_task.md
@@ -310,16 +310,6 @@ gym.register(
     "rl_cfg_entry_point": f"{__name__}.cartpole_env_cfg:RslRlOnPolicyRunnerCfg",
   },
 )
-
-gym.register(
-  id="Mjlab-Cartpole-Play",
-  entry_point="mjlab.envs:ManagerBasedRlEnv",
-  disable_env_checker=True,
-  kwargs={
-    "env_cfg_entry_point": f"{__name__}.cartpole_env_cfg:CartPoleEnvCfg",
-    "rl_cfg_entry_point": f"{__name__}.cartpole_env_cfg:RslRlOnPolicyRunnerCfg",
-  },
-)
 ```
 
 ---


src/mjlab/entity/entity.py
```diff
@@ -125,7 +125,8 @@ def __init__(self, cfg: EntityCfg) -> None:
       self._non_free_joints = tuple(all_joints[1:])
 
     self._apply_spec_editors()
-    self._add_initial_state_keyframe()
+    if not self._spec.key("init_state"):
+        self._add_initial_state_keyframe()
```
Gregwar: Note that this is a workaround and should be discussed, I might have misunderstood something
This is here because `key.ctrl = joint_pos` below seems to assume `ctrl` and `joint_pos` have the same size

kevinzakka:
Salut @Gregwar, it's great to see you use mjlab :)

I'm planning to do a refactor of actuators immediately after the sensor redesign (#206) lands. It will be spiritually similar: support built-in actuators (defined via XML or via config) and custom ones (e.g. `ActuatorNetMlp`). I believe this will resolve this issue and also properly handle hybrid setups (users who want to use both XML actuators and config-based ones). Do you think that's reasonable?"

Gregwar:
Hello @kevinzakka and thanks for your work!

It appear that you already support both XML and programmatically defined actuators, since this is what is happening (I guess by mistake) in this example

My remark is about the `ctrl` part of the `init_state` keyframe that should likely be implemented in another way: is that correct ?

louislelay:
Hi @Gregwar, you're right about that! It's currently being corrected in https://github.com/mujocolab/mjlab/pull/207.

Gregwar:
Another issue is that the `mdp.time_out` is not called in the example, it was replaced with `lambda env: False` for some reason (I also commited the change to keep trace, even if you don't merge it can act as a checklist)

(see e0fc4d8b0913363b481de77d050498e2894efef4)

Gregwar:
Also, because of the following:

```python
# From the CartPole example
@dataclass
  class CriticCfg(PolicyCfg):
    pass
```

And because the default configuration include both `policy` and `critic` for the `critic`:

```python
# src/mjlab/rl/config.py (RSL-RL default configuration)
default_factory=lambda: {"policy": ["policy"], "critic": ["policy", "critic"]},
```

The input of critic network becomes 8 (since critic input = "policy" + "critic" groups, where critic group equal policy)

Gregwar:
Also, the proposed reset for joints positions and velocities does nothing, as explained in #257

Gregwar:
I made some POC based on this example, but implemented in a separate repo using mjlab as a dependency here:

https://github.com/Gregwar/mjlab_cartpole

The environment setup can be hosted in a separate repo. For training & playing, small commands are defined importing relevant packages before calling mjlab commands:

https://github.com/Gregwar/mjlab_cartpole/blob/main/pyproject.toml#L15
https://github.com/Gregwar/mjlab_cartpole/blob/main/src/mjlab_cartpole/scripts/train.py

This could be an interesting pattern, instead of forking mjlab to add some tasks

louislelay:
Hello @Gregwar, that's what is being done by some people (including me) and I agree that this is the go-to way to do new tasks imo. I was actually thinking about doing a tutorial to explain how to set this up, wdyt?

https://github.com/mujocolab/g1_spinkick_example
https://github.com/BDX-R/BDX-R-MjLab
https://github.com/julien-blanchon/mjlab_goalkeeper

Gregwar:
Yes, I think this is a good idea to have such a thing in doc

Do you consider writing a readthedocs / sphinx for this repo ?

louislelay:
yes, i've actually opened the pr recently https://github.com/mujocolab/mjlab/pull/242, but we first need to merge some other things !

Gregwar:
Looks good!
We'll definitely dive into mjlab deeper and give feedbacks

kevinzakka:
Salut @Gregwar, how do you propose we merge your comments or repo example in this PR?

Gregwar:
@kevinzakka  I updated the PR, the CHANGELOG is:

1. Removing keyframe in XML that is not used
2. Using actuator from XML (i.e not declaring an extra one that is useless in the `.py`)
3. Fixing indent to 2
4. Fixing timeout by calling `mdp.time_out`
5. Fixing the `uv run play` command
6. Not registering `-Play` since it was actually the same env

Only the markdown is affected now.

About example repo, as @louislelay already said he wants to write some documentation about the setup of such repository. I can help re-writing this tutorial in that way if you want. Maybe this could be done in the RST branch directly

kevinzakka:
Thank you for all your efforts here! I'd love to get your help about the setup of such a repository. We could even put it in the mujocolab org repo if you're down?

kevinzakka:
Oh shoot @Gregwar, i would have love to see your gifs in the readme, they looked better!

Gregwar:
You mean this GIF from the repo? :slightly_smiling_face: 



The task is a little bit different from the tutorial (added a term to incentivize the cart to go to 0)

> I'd love to get your help about the setup of such a repository.

If @louislelay agree I can propose something

> We could even put it in the mujocolab org repo if you're down? 

Sure! Repositories mentionned by Louis use slightly different architecture, I don't have a really clear opinion about the details but it can be discussed in separate issue

PR #211, opened 2025-10-21
Add conda installation guide.

kevinzakka:
Also reflowed everything to 88 chars.

### DIFF
--- a/src/mjlab/entity/entity.py
+++ b/src/mjlab/entity/entity.py
@@ -158,7 +158,9 @@ class Entity:
     key = self._spec.add_key(name="init_state", qpos=key_qpos)
 
     if self.is_actuated and joint_pos is not None:
-      key.ctrl = joint_pos
+      name_to_pos = {name: joint_pos[i] for i, name in enumerate(self.joint_names)}
+      ctrl = np.array([name_to_pos.get(act.name, 0.0) for act in self._spec.actuators])
+      key.ctrl = ctrl
 
   # Attributes.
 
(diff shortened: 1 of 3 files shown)


docs/installation_guide.md
```diff
@@ -85,16 +90,60 @@ uv add mjlab "mujoco-warp @ git+https://github.com/google-deepmind/mujoco_warp@4
...
+
+1. **Create and activate your conda environment**:
+```bash
+conda create -n mjlab python=3.13
+conda activate mjlab
+```
+
+2. **Install PyTorch via conda**:
+```bash
+conda install pytorch pytorch-cuda=12.8 -c pytorch -c nvidia
```
louislelay: Were you actually able to install `pytorch-cuda=12.8` via conda? i’m getting `PackagesNotFoundError`. It seems the last conda build is 12.4 (see [pytorch-cuda on anaconda](https://anaconda.org/pytorch/pytorch-cuda)).

Isaac Lab gets 12.8 via the pip wheels instead (see their script: [https://github.com/isaac-sim/IsaacLab/blob/main/isaaclab.sh#L99](https://github.com/isaac-sim/IsaacLab/blob/main/isaaclab.sh#L99)), which boils down to something like:

```bash
pip install torch==2.7.0 torchvision==0.22.0 --index-url https://download.pytorch.org/whl/cu128
```

kevinzakka: Good catch. You actually don't need to manually install torch, confirmed by running the instructions myself. I also used the opportunity to generalize the instructions to any pip-based virtual environment.

PR #227, opened 2025-10-23
Fixes to NaN Guard

kevinzakka:
Previously, NaN guard only captured the first `max_envs_to_capture` environments (default 5). If a NaN occurred in environment 6+, it wouldn't be captured, making debugging impossible. We now capture all environments in the buffer and dump only the NaN environments (up to `max_envs_to_dump`).

2 extra QoL improvements:
* Creates `nan_dump_latest.npz` and `model_latest.mjb` for easy access without finding timestamps
* NaN visualizer now tracks the current environment, keeping robots centered when switching between environments

cc @rdeits-bd

### DIFF
--- a/tests/test_nan_guard.py
+++ b/tests/test_nan_guard.py
@@ -40,6 +40,7 @@ def test_nan_guard_disabled_by_default(simple_model):
   sim.close()
 
 
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="Likely bug on CPU MjWarp")
 def test_nan_guard_captures_and_dumps_on_nan(simple_model):
   """NaN guard should capture states and dump when NaN detected."""
   with tempfile.TemporaryDirectory() as tmpdir:


src/mjlab/scripts/nan_viz.py
```diff
@@ -199,14 +205,40 @@ def _update_state(self) -> None:
     mujoco.mj_setState(self.model, self.data, state, mujoco.mjtState.mjSTATE_PHYSICS)
     mujoco.mj_forward(self.model, self.data)
 
+    # Find the first non-fixed body and use its position for camera tracking.
```
rdeits-bd, line 208: Thank you! I was finding it super awkward to drag the viewer over to each environment.

rdeits-bd:
Verified that the dump works (and the `latest` symlink is great) -- I can train 4096 envs with `max-envs-to-dump = 1`, and it dumps exactly the 1 NaN env. 

However, I'm seeing a duplicate ground plane now in `viser` when using the `nan_viz`: 

image

kevinzakka:
@rdeits-bd Ugh, I only tested with a rough env. Let me fix that asap.

kevinzakka:
@rdeits-bd I just tried to reproduce and I only see 1 plane.

rdeits-bd:
Well that's interesting. I've restarted the viser tab as well but I still see the duplicate. I'll play around with it -- this isn't a blocker.

kevinzakka:
@rdeits-bd Don't worry about spamming me, I want to make sure this functionality is bullet proof and bug free. Can you try deleting the dump data, killing the viewer and trying again from scratch?

rdeits-bd:
Oh, this is probably just a silly mistake on my end -- I had a floor geometry in my robot xml.

Issue #230, opened 2025-10-24
Training not working out of the box due to missing onnxscript


Kukanani:
## Issue
I am able to run the `demo` without issues. However, training fails because `onnxscript` is missing.

Reproduction command (from README):
```bash
MUJOCO_GL=egl uv run train Mjlab-Velocity-Flat-Unitree-G1 --env.scene.num-envs 4096
```

Error (full error in complete output below):
```
  File "/home/adam/adam/1_projects/mjlab/.venv/lib/python3.13/site-packages/torch/onnx/_internal/exporter/_core.py", line 18, in <module>
    import onnxscript
ModuleNotFoundError: No module named 'onnxscript'
```

## Fix Attempts
* I've checked `onnx` and `rsl_rl` repositories for this issue and didn't find anything.
* If I `uv add onnxscript`, training does not crash and loss goes down. However, the script continuously throws errors on every 50th iteration. Perhaps these can be safely ignored?
```
/home/adam/adam/1_projects/mjlab/src/mjlab/third_party/isaaclab/isaaclab_rl/rsl_rl/exporter.py:204: UserWarning: # 'dynamic_axes' is not recommended when dynamo=True, and may lead to 'torch._dynamo.exc.UserError: Constraints violated.' Supply the 'dynamic_shapes' argument instead if export is unsuccessful.
  torch.onnx.export(
W1023 22:04:44.594000 3118709 torch/onnx/_internal/exporter/_compat.py:114] Setting ONNX exporter to use operator set version 18 because the requested opset_version 11 is a lower version than we have implementations for. Automatic version conversion will be performed, which may not be successful at converting to the requested version. If version conversion is unsuccessful, the opset version of the exported model will be kept at 18. Please consider setting opset_version >=18 to leverage latest ONNX features
The model version conversion is not supported by the onnxscript version converter and fallback is enabled. The model will be converted using the onnx C API (target version: 11).
Failed to convert the model to the target version 11 using the ONNX C API. The model was not modified
Traceback (most recent call last):
  File "/home/adam/adam/1_projects/mjlab/.venv/lib/python3.13/site-packages/onnxscript/version_converter/__init__.py", line 127, in call
    converted_proto = _c_api_utils.call_onnx_api(
        func=_partial_convert_version, model=model
    )
  File "/home/adam/adam/1_projects/mjlab/.venv/lib/python3.13/site-packages/onnxscript/version_converter/_c_api_utils.py", line 65, in call_onnx_api
[...]

PR #235, opened 2025-10-24
Support exporter for both pytorch 2.9 and backward

MiaoDX:
As discussed in https://github.com/mujocolab/mjlab/issues/230

Maybe we should update the codes for the long run?

Tested Velocity task on `2.7.0+cu128 and 2.9.0+cu128`, but it seems Tracking task is not so easy to train directly?

Note, not meant to merge directly, since the training phase is not done, I am not so sure whether this actually works or not, all codes are done by codex.

But just raise one question, do we need to upgrade all the libs and behave as needed in the long run.


kevinzakka:
Addressed by #250.

PR #239, opened 2025-10-24
Add RSL-RL teacher-student distillation support

### LINKED ISSUE #208: Adding support for the student-teacher distillation in RSL-RL
Hi @kevinzakka,

thank you for the great (as always) project! I would really like to see this come out of Beta and into Stable, and am interested in contributing to make sure that happens.

I imagine student-teacher distillation is part of #100. It also happens to be something I currently need. If you haven't started that yet, happy to give it a shot. Would a PR for the `Isaac-Velocity-Flat-Anymal-D-v0` distillation task from the [IsaacLab docs](https://isaac-sim.github.io/IsaacLab/main/source/overview/reinforcement-learning/rl_existing_scripts.html#rsl-rl) be a good way to go about this?

jaraujo98:
Closes #208. Changes:

* Added the configuration classes for the RSL-RL teacher-student distillation algorithm, model, and runner.
* Edited the train and play scripts to support this.
* Added a velocity tracking task with the UnitreeGo1 similar to what was available in IsaacLab (though when I went to double check I realized the task I added is not exactly the same, but it still works to illustrate the functionality).

I tested this by training the policies as described on the README.


src/mjlab/tasks/velocity/config/go1/rl_cfg.py
```diff
@@ -39,3 +42,32 @@ class UnitreeGo1PPORunnerCfg(RslRlOnPolicyRunnerCfg):
   save_interval: int = 50
   num_steps_per_env: int = 24
   max_iterations: int = 10_000
+
+
+@dataclass
+class UnitreeGo1DistillRunnerCfg(RslRlDistillationRunnerCfg):
```
louislelay:
```suggestion
class UnitreeGo1DistillationRunnerCfg(RslRlDistillationRunnerCfg):
```

src/mjlab/tasks/velocity/config/go1/rl_cfg.py
```diff
@@ -39,3 +42,32 @@ class UnitreeGo1PPORunnerCfg(RslRlOnPolicyRunnerCfg):
...
+  num_steps_per_env: int = 120
+  max_iterations: int = 10_000
+  save_interval: int = 500
+  experiment_name: str = "go1_velocity"
+  obs_groups: dict[str, list[str]] = field(
```
louislelay: NIT: Could you place it at the end in a similar fashion as the other config?

louislelay:
Hello @jaraujo98, thank you very much! We discussed it and think it’s best to keep the repository as slim as possible. I’d recommend creating an external repository for this and sharing it in the discussions. Ultimately, we’d like to add documentation on how to integrate RL libraries/algorithms so people can implement the ones they need!

jaraujo98:
Thank you for having a look, @louislelay! I've been hacking with `mjlab` a lot, and at this point I agree 100%. I'll close this and the original issue.

PR #245, opened 2025-10-27
Add plugin system with auto-subscription

julien-blanchon:
This PR draft add a plugin system with auto-subscription, basically any library could declare new gym using project entrypoint (https://packaging.python.org/en/latest/specifications/entry-points/). And they will be automatically added to the mjlab env list without having to wrap the play/train function or anything.

Here is how it work

1. In the extension library add some entrypoint for the `mjlab.envs` in the pyproject.toml. See this example in [mjlab_goalkeeper](https://github.com/julien-blanchon/mjlab_goalkeeper), where mjlab_goalkeeper/__init__.py is exporting some gym (https://github.com/julien-blanchon/mjlab_goalkeeper/blob/main/src/mjlab_goalkeeper/__init__.py)

```toml
[project.entry-points."mjlab.envs"]
mjlab-goalkeeper = "mjlab_goalkeeper"
```

2. Those entrypoint will be automated loaded in the mjlab __init__.py (see https://github.com/julien-blanchon/mjlab/blob/a6b30e313a7be8fdc72aad90bfb0e6fd9ab94fac/src/mjlab/__init__.py#L19-L28), and be available everywhere.


julien-blanchon:
Hey guys any interest in merging this ?

kevinzakka:
Hi @julien-blanchon,

What do you think of something [like this](https://github.com/mujocolab/g1_spinkick_example/blob/main/train.py)?

julien-blanchon:
The goal is expecially to avoid this for multiple reason:
- You need to explicitly add those play.py, run.py, list_envs.py boilerplate to every new project
- It's quite counter intuitive to require to import the new env lib first with `# noqa: F401` ...
- It's broking script use (`uv run train/play ...`)
- It could be very quickly a mess if you want to work with multiple new environments

kevinzakka:
Hi @julien-blanchon, I actually didn't know about this plugin system business. I did some local testing and read about it and really like it. Since I need to add some documentation + change the g1_spinkick_example repo, I'm going to create a PR and credit you there. Thanks again for the great suggestion!

PR #236, opened 2025-10-24
Add docker container and docs

Kukanani:
As requested in #230 

(note that this does not resolve the issue in #230, the docker container throws the same error for now)

Pushing to Docker Hub is left as an exercise for the maintainer

@kevinzakka @MiaoDX

PR #249, opened 2025-10-27
Add helper script to run docker with appropriate flags and reference in the doc

Kukanani:
follow-up on #236 , builds on that one.

Patterned on the `uv` in docker example. I changed one thing, which was fixing the tag (name) of the image to `mjlab` so it can be referenced by other tools after build.

### DIFF
--- a/scripts/run_docker.sh
+++ b/scripts/run_docker.sh
@@ -1,19 +1,22 @@
 #!/usr/bin/env sh
 #
-# Injects useful arguments for running mjlab in docker. See docs/installation_guide.md for usage.
-# This is patterned after the uv-in-docker example: https://github.com/astral-sh/uv-docker-example/blob/5748835918ec293d547bbe0e42df34e140aca1eb/run.sh
+# Injects useful arguments for running mjlab in docker.
+# See docs/installation_guide.md for usage.
 #
-# docker run \
-#     --rm \                        Remove the container after exiting
-#     --runtime=nvidia \            Use NVIDIA Container runtime to give GPU access
-#     --gpus all \                  Expose all GPUs by default
-#     --volume .:/app \             (follows uv example) Mount the current directory to `/app` so code changes don't require an image rebuild
-#     --volume /app/.venv \         (follows uv example) Mount the virtual environment separately, so the developer's environment doesn't end up in the container
-#     --publish 8080:8080 \         publish port 8080 for viewing the mjlab web interface on the host
-#     $INTERACTIVE \                if in a terminal, launch in interactive mode. Note that if running training,
-#                                   there is a blocking wandb prompt before training will begin.
-#     $(docker build -t mjlab) \    build and launch the docker image (tag matches the one in the Makefile)
-#     "$@"                          forward all arguments
+# Patterned after the uv-in-docker example:
+# https://github.com/astral-sh/uv-docker-example/blob/5748835918ec293d547bbe0e42df34e140aca1eb/run.sh
+#
+# Key arguments:
+#   --rm                Remove the container after exiting
+#   --runtime=nvidia    Use NVIDIA Container runtime to give GPU access
+#   --gpus all          Expose all GPUs by default
+#   -v .:/app           Mount current directory to /app (code changes don't require rebuild)
+#   -v /app/.venv       Mount venv separately (keeps developer's environment out of container)
+#   -p 8080:8080        Publish port 8080 for viewing mjlab web interface on the host
+#   -it (conditional)   Launch in interactive mode if running in a terminal
+#                       (Note: if running training, there's a blocking wandb prompt before training begins)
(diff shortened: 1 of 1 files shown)


run_docker.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env sh
```
kevinzakka: Can you create a scripts/ dir and move this file there?

run_docker.sh
```diff
@@ -0,0 +1,34 @@
...
+#     --rm \                        Remove the container after exiting
+#     --runtime=nvidia \            Use NVIDIA Container runtime to give GPU access
+#     --gpus all \                  Expose all GPUs by default
+#     --volume .:/app \             (follows uv example) Mount the current directory to `/app` so code changes don't require an image rebuild
+#     --volume /app/.venv \         (follows uv example) Mount the virtual environment separately, so the developer's environment doesn't end up in the container
+#     --publish 8080:8080 \         publish port 8080 for viewing the mjlab web interface on the host
+#     $INTERACTIVE \                if in a terminal, launch in interactive mode. Note that if running training,
+#                                   there is a blocking wandb prompt before training will begin.
+#     $(docker build -t mjlab) \    build and launch the docker image (tag matches the one in the Makefile)
+#     "$@"                          forward all arguments
```
kevinzakka: Mind switching to something like this?

```bash
#!/usr/bin/env sh
#
# Injects useful arguments for running mjlab in docker.
# See docs/installation_guide.md for usage.
# Patterned after the uv-in-docker example:
# https://github.com/astral-sh/uv-docker-example/blob/5748835918ec293d547bbe0e42df34e140aca1eb/run.sh
#
# Key flags:
#   --rm                Remove the container after exiting
#   --runtime=nvidia    Use NVIDIA Container runtime to give GPU access
#   --gpus all          Expose all GPUs by default
#   -v .:/app           Mount current directory to /app (code changes don't require rebuild)
#   -v /app/.venv       Mount venv separately (keeps developer's environment out of container)
#   -p 8080:8080        Publish port 8080 for viewing mjlab web interface on the host
#   -it (conditional)   Launch in interactive mode if running in a terminal
#                       (Note: if running training, there's a blocking wandb prompt before training begins)
#   docker build        Build and launch the image (tag matches the Makefile)
#   "$@"                Forward all arguments to the docker image

if [ -t 1 ]; then
    INTERACTIVE="-it"
else
    INTERACTIVE=""
fi

docker run \
    --rm \
    --runtime=nvidia \
    --gpus all \
    --volume .:/app \
    --volume /app/.venv \
    --publish 8080:8080 \
    $INTERACTIVE \
    $(docker build -qt mjlab .) \
    "$@"
```

kevinzakka (changes requested):
Minor nit, other than than LGTM! Thanks for doing this so quickly!

run_docker.sh
kevinzakka: Unless you feel strongly about this. I typically like scripts in such a folder.

docs/installation_guide.md
```diff
@@ -137,9 +137,9 @@ also use it with any pip-based virtual environment (venv, conda, virtualenv, etc
 - Install an appropriate NVIDIA driver for your system and the [NVIDIA Container Toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html)
   - Be sure to register the container runtime with Docker and restart, as described in [the Docker section of the install guide](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html#configuring-docker)
 - `make docker-build`
-- `docker run --rm --runtime=nvidia mjlab:latest`
-- Demo with viewer: `docker run --rm -it -p 8080:8080 --runtime=nvidia --gpus all mjlab:latest uv run demo`
-- Training example:`docker run --rm -it -p 8080:8080 --runtime=nvidia --gpus all mjlab:latest uv run train Mjlab-Velocity-Flat-Unitree-G1 --env.scene.num-envs 4096`
+- `./run_docker.sh`
```
kevinzakka: Don't forget `./scripts/run_docker.sh`

run_docker.sh
Kukanani: done. It's all rewording for brevity and clarity, I think?

100% curiosity question - is this your personal writing style or AI cleanup?

run_docker.sh
Kukanani: moved

run_docker.sh
kevinzakka: Oh you totally don't have to change the language. I just thought it would be clearer this way and not get screwed up due to line wrapping.

In this specific instance I asked claude to give me a docstring at the top in a more formal shell style lol.

kevinzakka:
Thanks again @Kukanani!

PR #250, opened 2025-10-28
Fixes for PyTorch 2.9.

kevinzakka:
Fixed a few issues and warnings from the torch upgrade to 2.9.

- The `rsl_rl` exporter logic in the third_party file now explicitly sets `dynamo=False`. This was previously the default in torch<2.9. This is probably not ideal in the long-run but a good short-term solution.
- `configure_torch_backends` now uses the new fp32_precision semantics. This gets rid of the annoying warning we see when running train and play scripts.

### DIFF
--- a/src/mjlab/third_party/isaaclab/README.vendor.md
+++ b/src/mjlab/third_party/isaaclab/README.vendor.md
@@ -4,5 +4,12 @@
 - **License:** BSD-3-Clause (see [LICENSE](LICENSE))
 - **Commit/tag:** `5f71ff479eb121d8aff6c37caaf6768927e1e5c9`
 
-We include only these files as-is to support our code. For full source,
+## Local Modifications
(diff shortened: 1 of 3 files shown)


kevinzakka:
cc @Kukanani

pyproject.toml
```diff
@@ -46,6 +46,7 @@ dependencies = [
     "tensordict",
     "rsl-rl-lib",
     "tensorboard>=2.20.0",
+    "onnxscript>=0.5.4",
```
Kukanani, line 49: oh, now I'm curious where you got v0.5.4. PyTorch pins 0.4.0, but obviously if it can use newer versions that seems preferable

https://github.com/pytorch/pytorch/blob/v2.9.0/.ci/docker/common/install_onnx.sh#L23

PR #258, opened 2025-10-29
Fix velocity task. 

kevinzakka:
I tried to improve the velocity tasks with the sensor API redesign, when I should have done them in 2 separate PRs. In the process of doing so, I broke a bunch of things in the velocity tasks which I'm addressing here.

Really happy with the final result. There are a few small things I'd improve still but I think this is a good stopping point. Posting sim and real videos below.

https://github.com/user-attachments/assets/e8dabbbc-af89-4e47-9aba-db87233b3a80

https://github.com/user-attachments/assets/f1f8c030-78fe-4d5a-9a15-f46b648a7acc

Note for future Kevin: investigate the `soft_landing` reward. It looks like it's causing the ankles to roll in a weird way and I suspect there might be a weird bug in the net wrench calculation.

cc @yuvaltassa


Gregwar:
Looks really good! 🔥

PR #266, opened 2025-10-30
Add generic curriculum for changing reward weight

StafaH:
Adds a general reward weight curriculum similar to IsaacLab.


kevinzakka:
It's already in #258.

StafaH:
Cool, can we move it outside of the velocity task into the main mdp?

PR #281, opened 2025-11-03
Replace reset_joints_by_scale with reset_joints_by_offset.

### LINKED ISSUE #257: Why `reset_joints_by_scale` ?
In the function `reset_joints_by_scale` of [events.py](https://github.com/mujocolab/mjlab/blob/fccf5470ed899a6d028a0367d6185d2ffd6deee7/src/mjlab/envs/mdp/events.py#L158), joints positions and velocity are computed this way:

```python
  joint_pos = default_joint_pos[env_ids][:, asset_cfg.joint_ids].clone()
  joint_vel = default_joint_vel[env_ids][:, asset_cfg.joint_ids].clone()

  joint_pos *= sample_uniform(*position_range, joint_pos.shape, env.device)
  joint_vel *= sample_uniform(*velocity_range, joint_vel.shape, env.device)
```

I don't understand the underlying logic in multiplying default joint positions with a scalar. I'd expect most joints to have a default of 0 with a default velocity of also 0 and this method to do nothing (this is what happen in my current tests), which is confusing.

Shouldn't it be something like (pseudo-code):

```python
joint_pos = default_joint_pos + uniform * joint_range
```

Where `joint_range` would express the amplitude of the joint ?
(Not sure about what that would mean for velocities though since there is no velocity limits)

kevinzakka:
Fixes #257.

cc @Gregwar

### DIFF
--- a/docs/create_new_task.md
+++ b/docs/create_new_task.md
@@ -236,7 +236,7 @@ def random_push_cart(env, env_ids, force_range=(-5, 5)):
 class EventCfg:
   reset_robot_joints: EventTerm = term(
     EventTerm,
-    func=mdp.reset_joints_by_scale,
+    func=mdp.reset_joints_by_offset,
     mode="reset",
     params={
       "asset_cfg": SceneEntityCfg("robot"),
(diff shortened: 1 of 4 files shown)


Gregwar:
Looks good to me!

Issue #270, opened 2025-10-30
Make project creation easier


louislelay:
Hello @Gregwar and @kevinzakka, 

I just saw your latest comments on PR #209. Since it’s now closed, I thought I’d open an issue to continue the discussion.

I was planning to add this tutorial idea directly to my branch with the sphinx docs, but if you'd like to propose one for the markdown docs in the meantime, that works for me!

Fyi without entering too much into the details, this is what I typically do:

```bash
uv init --package foo_mjlab && cd foo_mjlab
uv add mjlab "git+https://github.com/mujocolab/mjlab"
```

Then I add scripts that subclass the mjlab scripts, and I add the assets/tasks in:

```
src/foo_mjlab/robots/

src/foo_mjlab/tasks/
```

My initial idea was to document this, but I’m wondering if it might be cleaner to make Grégoire's repo a github template in the org.

Also, Grégoire fyi: there was also a script in progress at one point to create external projects in a similar fashion to Isaac Lab.

I'd love to get thoughts from both of you on the best path forward. Looping in @brentyi since we discussed this a bit before the release!

PS: Sorry for the typos, I'm on my phone.

PR #282, opened 2025-11-03
Add plugin system for external task registration.

kevinzakka:
Enables external plugins to register custom tasks via Python entry points. Plugins are automatically discovered and loaded at runtime, making them available to train/play scripts.

### DIFF
--- a/docs/create_new_task.md
+++ b/docs/create_new_task.md
@@ -333,3 +333,37 @@ uv run play Mjlab-Cartpole --checkpoint_file <checkpoint_path>
 ```
 
 ![Trained CartPole](static/cartpole_trained.gif)
+
+---
+
+## 5. External Package Registration (Optional)
+
+To develop your task in a separate repository outside of mjlab, add an entry
+point to your package's `pyproject.toml`:
+
+> [!NOTE]
+> The `[build-system]` section is required for the plugin system to work.
+
+```toml
+[build-system]
+requires = ["uv_build>=0.8.19,<0.9.0"]
+build-backend = "uv_build"
+
+[project.entry-points."mjlab.tasks"]
+your_package_name = "your_package_name"
+```
+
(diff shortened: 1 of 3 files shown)


kevinzakka:
@julien-blanchon Let me know if this looks good to you!

kevinzakka:
@louislelay I believe this resolves #270?

docs/create_new_task.md
```diff
@@ -333,3 +333,37 @@ uv run play Mjlab-Cartpole --checkpoint_file <checkpoint_path>
 ```
 
 ![Trained CartPole](static/cartpole_trained.gif)
+
+---
+
+## 5. External Package Registration (Optional)
+
+To develop your task in a separate repository outside of mjlab, add an entry
+point to your package's `pyproject.toml`:
```
brentyi, line 342: should we add a note about targeting specific `mjlab` versions? (since currently there are no backwards compatibility guarantees)

kevinzakka: Chatted offline, will address more thoroughly once we exit beta.

louislelay:
Yes it does! Thanks @kevinzakka 

Edit: in that case, I think https://github.com/mujocolab/mjlab/pull/279 make even more sense to be merged, wdyt? Because if we run `list_envs`, it will list all the environments from all external projects on the computer.

julien-blanchon:
> @julien-blanchon Let me know if this looks good to you!

Yes good for me ! I'm sad my PR was closed tho xd

> Because if we run list_envs, it will list all the environments from all external projects on the computer.

I also think this would make sence. A lot of things could be added on top of this: tag system, versionning ...
But it's maybe better to merge this before, and later think about improving on top

julien-blanchon:
About https://github.com/mujocolab/mjlab/issues/270, I think a cookiecutter with a a minimal task example would make sense.

And `uv tool run cookiecutter ...`

PR #288, opened 2025-11-06
Add `domain_randomization` field to `EventTermCfg` to replace `__name__` inspection. 

### LINKED ISSUE #268: Event terms as class and randomization fields
If one want to use a class for an event term, just like the below example, the following line of code will raise an error:
https://github.com/mujocolab/mjlab/blob/89fdd41e210d3e42c1117593ac4147da3ae46926/src/mjlab/managers/event_manager.py#L210

Because `__name__` is required and does not exist within a class. 

Moreover, this implementation is arguably brittle since it doesn't let user easily add some events that would imply domain randomization

## Proposition (to be discussed)

One possibility would be to make all the events that imply randomizing field classes that inherit a known abstract class. This way the event manager could detect them with an instance test. What do you think ?

## Example

```python
class randomize_terrain_friction:
  def __init__(self, cfg, env):
    self._terrain_idx = None
    for idx, geom in enumerate(env.scene.spec.geoms):
      if geom.name == "terrain":
        self._terrain_idx = idx

    if self._terrain_idx is None:
      raise ValueError("Terrain geom not found in the model.")

  def __call__(self, env, env_ids, ranges):
    env.sim.model.geom_friction[env_ids, self._terrain_idx, 0] = sample_uniform(
      ranges[0], ranges[1], len(env_ids), env.device
    )

  # Possible workaround, adding __name__ inside the class
  __name__ = "randomize_terrain_friction"
```

kevinzakka:
Previously, the event manager relied on checking `func.__name__ == "randomize_field"` to detect domain randomization terms. This broke when users tried to use class-based event terms.

This adds an explicit `domain_randomization` boolean field to EventTermCfg that users can set to mark DR terms. This approach is more explicit, works with any callable (functions, classes, lambdas), and makes the intent clear in the configuration.

Fixes #268.

### DIFF
--- a/docs/api/domain_randomization.md
+++ b/docs/api/domain_randomization.md
@@ -21,6 +21,7 @@ foot_friction: EventTerm = term(
     EventTerm,
     mode="reset",  # randomize each episode
     func=mdp.randomize_field,
+    domain_randomization=True,  # marks this as domain randomization
     params={
         "asset_cfg": SceneEntityCfg("robot", geom_names=[".*_foot.*"]),
         "field": "geom_friction",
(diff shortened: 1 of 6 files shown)


kevinzakka:
@rdeits-bd Let me know if this LGTY?

rdeits-bd (approved):
Looks great!

PR #292, opened 2025-11-07
[Breaks existing configs] Fewer classes in configs

brentyi:
Reimplementation of #124 to avoid dealing with merge conflicts.

---

This PR reduces the number of classes we use in the config system:
- As key-value stores. We define dataclasses where each field is some kind of action/observation/reward/etc term, and then call a `get_terms()` helper to list out terms.
- For config inheritance. A Go1 velocity config inherits some fields from a base velocity config, which inherits some fields from the base ManagerBasedRlEnvConfig.
    - Some overrides are done by overriding dataclass fields.
    - Other overrides are done in `__post_init__()`.

In this PR, we:
- Switch to dictionaries whenever key-value pairs are needed. 
- Flatten the config hierarchy. All configs now directly use `ManagerBasedRlEnvConfig`. 

I think this is more reasonable than needing to define lots of new classes for each new task, but it will be less familiar for folks who are used to IsaacLab. Interested in thoughts!

## Why?

The main change here is to replace specialized config _classes_ with config _instances_. There are multiple reasons why this makes sense to me:

### Current configs that rely on `__post_init__` mutations are fundamentally broken.

A pattern that both `IsaacLab` and `mjlab` adopt right now is:
```python
@dataclass
class SomeBaseEnvCfg:
    episode_length_s: float = 5.0
    # ...and other fields

@dataclass
class NewTaskEnvCfg(SomeBaseEnvCfg):
    def __post_init__(self) -> None:
        self.episode_length_s = 20.0
```

The problem with this is that config values passed into the class constructor will be overridden:
```python
cfg = NewTaskEnvCfg(episode_length_s=10.0)    # Set value to 10.0.
print(cfg.episode_length_s)                   # 20.0 because of __post_init__().
```

This is unintuitive and also applies to config values passed in the CLI; in `mjlab`, `--env.episode-length-s 10.0` would be ignored for the example above. Instance-based configs don't have this problem.

### The alternative fix is unergonomic and error-prone.

To fix the broken `__post_init__` behavior, a common thing the current code does is to redeclare fields that we want to change the default value for:
```python
@dataclass
class NewTaskEnvCfg(SomeBaseEnvCfg):
    episode_length_s: float = 20.0
```
The behavior is now correct, but overriding nested config values this way is tedious. It also feels like an invitation for bugs, since we don't have a way (both statically and at runtime) to verify that the new declaration is actually overriding an existing field vs creating a new one. Renaming the base config's field from `episode_length_s` to `episode_length_sec` will produce silent errors.

Instance-based configs don't have these problems. Nested mutation of configs is simple; there's also less room for silent errors when we set fields via constructor args `ManagerBasedRlEnvCfg(..., episode_length_s=20.0)` or mutation `cfg.episode_length_s = 20.0`.

### Simple things become simpler.

Let's say we want to create a minor variation of an existing task, with an extra observation term for the critic network. This currently requires several new classes: for the task, for the observations, for the critic observation group. All of these classes will need field overrides that will fail silently if we misspell `observations`, write `value` instead of `critic`, etc.  

Life is simpler with instance-based configs; we can simply copy the original task config and modify it:
```python
new_task = deepcopy(original_task)

# Get critic observation terms.
new_task_critic_terms = new_task.observations["critic"].terms

# Add a new term.
assert "new_term" not in new_task_critic_terms
new_task_critic_terms["new_term"] = ObservationTermCfg(...)
```

### Type safety + self-documentation improves.

The "use dataclasses as key-value stores" config approach results in code that's littered with `Any` and `object`. Fields like `rewards: dict[str, RewardTermCfg]` feel much clearer to me.

### DIFF
--- a/src/mjlab/envs/mdp/events.py
+++ b/src/mjlab/envs/mdp/events.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Dict, List, Literal, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Dict, Literal, Tuple, Union
 
 import torch
 
@@ -221,8 +221,8 @@ class FieldSpec:
 
   entity_type: Literal["dof", "joint", "body", "geom", "site", "actuator"]
   use_address: bool = False  # True for fields that need address (q_adr, v_adr)
-  default_axes: Optional[List[int]] = None
-  valid_axes: Optional[List[int]] = None
+  default_axes: list[int] | None = None
+  valid_axes: list[int] | None = None
 
 
 FIELD_SPECS = {
@@ -261,7 +261,7 @@ def randomize_field(
   distribution: Literal["uniform", "log_uniform", "gaussian"] = "uniform",
   operation: Literal["add", "scale", "abs"] = "abs",
   asset_cfg=None,
-  axes: Optional[List[int]] = None,
+  axes: list[int] | None = None,
 ):
   """Unified model randomization function.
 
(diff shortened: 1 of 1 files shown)


kevinzakka (approved):

kevinzakka:
Thanks @brentyi!

louislelay:
Thanks a lot @brentyi!

PR #297, opened 2025-11-08
Fix frame mismatch in root velocity write operations.

kevinzakka:
`write_root_velocity` now expects angular velocity in world frame (matching `root_link_vel_w`), and internally converts to body frame for qvel. This ensures `write_root_velocity(root_link_vel_w) `is a no-op.

Updated call sites in events.py, commands.py, and csv_to_npz.py to remove manual `quat_apply_inverse` conversions. Added tests for round-trip consistency and frame conversion.

### DIFF
--- a/tests/test_entity_data.py
+++ b/tests/test_entity_data.py
@@ -42,46 +42,55 @@ def initialize_entity_with_sim(entity, device, num_envs=1):
 
 
 def test_root_velocity_world_frame_roundtrip(device):
-  """Test that reading and writing root velocity is a no-op (both in world frame)."""
+  """Test reading and writing root velocity is a no-op (world frame).
+
+  Verifies that the API is consistent: if you write a velocity, read it
+  back, and write it again, you get the same result. This ensures no
+  unintended transformations happen in the read/write cycle.
+  """
   entity = create_floating_base_entity()
   entity, sim = initialize_entity_with_sim(entity, device)
 
-  # fmt: off
-  root_state = torch.tensor([
-    0.0, 0.0, 1.0,
-    0.6, 0.2, 0.3, 0.7141,
-    1.0, 0.5, 0.0,
-    0.0, 0.3, 0.1
-  ], device=device).unsqueeze(0)
-  # fmt: on
-  entity.write_root_state_to_sim(root_state)
+  pose = torch.tensor([0.0, 0.0, 1.0, 0.6, 0.2, 0.3, 0.7141], device=device).unsqueeze(
+    0
+  )
+  entity.write_root_link_pose_to_sim(pose)
+
+  vel_w = torch.tensor([1.0, 0.5, 0.0, 0.0, 0.3, 0.1], device=device).unsqueeze(0)
+  entity.write_root_link_velocity_to_sim(vel_w)
   sim.forward()
 
-  vel_w_before = entity.data.root_link_vel_w.clone()
-  entity.write_root_link_velocity_to_sim(vel_w_before)
+  vel_w_read = entity.data.root_link_vel_w.clone()
+  assert torch.allclose(vel_w_read, vel_w, atol=1e-4)
+
+  entity.write_root_link_velocity_to_sim(vel_w_read)
   sim.forward()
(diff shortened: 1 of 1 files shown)


kevinzakka:
cc @rdeits-bd

src/mjlab/entity/data.py
```diff
@@ -94,7 +94,13 @@ def write_root_velocity(
...
 
     env_ids = self._resolve_env_ids(env_ids)
-    self.data.qvel[env_ids, self.indexing.free_joint_v_adr] = velocity
+    lin_vel_w = velocity[:, :3]
+    ang_vel_w = velocity[:, 3:]
+    qpos_free = self.data.qpos[env_ids, self.indexing.free_joint_q_adr]
+    quat_w = qpos_free[:, 3:7]
+    ang_vel_b = quat_apply_inverse(quat_w, ang_vel_w)
+    velocity_qvel = torch.cat([lin_vel_w, ang_vel_b], dim=-1)
+    self.data.qvel[env_ids, self.indexing.free_joint_v_adr] = velocity_qvel
```
rdeits-bd, line 106: What do you think about making this explicit in the function name rather than changing the behavior of the existing function? That is, while you're changing the behavior also change the name to `write_root_velocity_w` ? That would catch any uses of the old function in downstream code rather than potentially silently changing behavior.

kevinzakka: Yeah that seems better. I did this because technically speaking the documentation was misleading. 

I did a few tests since you opened the issue. 1) I tried some defensive `sim.forward()` everywhere in entity.data to guaranteee freshness. That killed perf on the tracking task. 2) I tried having a debug mode that tracks writes to qpos/qvel/act/xfrc_applied/qfrc_applied and then checks if derived properties are read after these "dirty" writes and warns the user but it also didn't feel like the right approach. 

Stepping back a bit, I definitely don't want to have a library that is silently doing things without the user's knowledge. `dm_control` used to have similar behavior and it caused all sorts of "gotchas" for downstream users.

I think our best bet is to document things better (e.g. order of events matters), fix this root velocity thing (since it will be jarring for isaac users) and then also tell users it is their responsibility to call sim.forward() when they modify the state and then try to access some derived property later.

rdeits-bd: Yeah I think that's fine as a philosophy. One thing that made IsaacLab's performance hard to reason about was that a lot of operations silently did quite a lot of work in the background, and it was tricky to figure out what those were.

A couple other thoughts:

* Could the entity data at least *know* which environments have dirty forward kinematics state? I'm not actually sure -- the `qpos` and `qvel` are just views into the internal mujoco_warp data, so it's not obvious to me if you'd be able to provide an `is_xpos_valid()` function or not. 
* I agree that extra `sim.forward()` calls tank performance. I think a big part of that, though, is that `forward()` operates on all environments, so if you have an event that wants updated FK data you're going to pay that cost for all environments when just a single environment has changed. It doesn't seem like mujoco warp exposes a per-world forward call, though.

kevinzakka: Hi Robin, mjwarp does not have a per-world forward call unfortunately (and it might not necessarily be that much more perf-friendly than the full one). Here's what I've decided with respect to the API:

1. All write methods write to the state (qpos, qvel, xfrc_applied, ctrl, etc.) Thus write methods are all "up-to-date" in that sense.
2. Read properties, on the other hand, read from derived quantities (e.g., xpos, cvel) and thus, it is the user's responsibility to call sim.forward() if they need up-to-date values with respect to the current state. This is particularly important when using multiple events: if Event A writes state and Event B reads state, you must call sim.forward() between them.
3. All inputs and outputs use world frame. Internal MuJoCo conventions (e.g., angular velocity in body frame) are handled automatically.

kevinzakka: Let me know if the [new tests](https://github.com/mujocolab/mjlab/blob/ed3e5a5de49002d13b9a683905837530fcf57820/tests/test_entity_data.py) I added make sense!

bd-aadubredu:
thank you @kevinzakka !

Issue #301, opened 2025-11-10
Broken list_envs script due to recent changes


louislelay:
When I run list_envs, the CLI prints a massive table where the Env Cfg Entry Point column contains the entire ManagerBasedRlEnvCfg(...) dataclass repr for each environment. This make the output extremely wide

PR #303, opened 2025-11-10
Modifies env-cfg entry points to string so list_envs prints compact environment table

louislelay:
This PR switches `env_cfg_entry_point` to string-based entry points.

Previously, environments were registered like this:

```python
kwargs = {
  "env_cfg_entry_point": UNITREE_G1_FLAT_TRACKING_ENV_CFG,
  ...
}
```

We had short repr strings like `<class 'module.G1FlatEnvCfg'>`, so `list_envs` displayed them compactly.

But since we refactored these configs from `@dataclass` classes to `@retval` functions. Now `G1_FLAT_TRACKING_ENV_CFG` is a fully instantiated `ManagerBasedRlEnvCfg` object. When `list_envs` prints this, PrettyTable renders the entire nested object repr, making the table unreadable:

```text
| Env Cfg Entry Point                                                                                              |
| ManagerBasedRlEnvCfg(scene=SceneCfg(...), actions=ActionsCfg(...), ...) |
```

With this PR, `list_envs` always sees a `str` for `env_cfg_entry_point` and just prints it. The reported type is now `builtins:str`, and the table output is compact and readable again:

```text
+-------------------------------------------------------------------------------------------------------------------------------------------------------+
|                                                            Available Environments in Mjlab                                                            |
+---+----------------------------------------------------+----------------------------------------------------------------------------------------------+
| # | Task ID                                            | Env Cfg Entry Point                                                                          |
+---+----------------------------------------------------+----------------------------------------------------------------------------------------------+
| 1 | Mjlab-Tracking-Flat-Unitree-G1                     | mjlab.tasks.tracking.config.g1.env_cfgs:UNITREE_G1_FLAT_TRACKING_ENV_CFG                     |
| 2 | Mjlab-Tracking-Flat-Unitree-G1-No-State-Estimation | mjlab.tasks.tracking.config.g1.env_cfgs:UNITREE_G1_FLAT_TRACKING_NO_STATE_ESTIMATION_ENV_CFG |
| 3 | Mjlab-Velocity-Rough-Unitree-G1                    | mjlab.tasks.velocity.config.g1.env_cfgs:UNITREE_G1_ROUGH_ENV_CFG                             |
| 4 | Mjlab-Velocity-Flat-Unitree-G1                     | mjlab.tasks.velocity.config.g1.env_cfgs:UNITREE_G1_FLAT_ENV_CFG                              |
| 5 | Mjlab-Velocity-Rough-Unitree-Go1                   | mjlab.tasks.velocity.config.go1.env_cfgs:UNITREE_GO1_ROUGH_ENV_CFG                           |
| 6 | Mjlab-Velocity-Flat-Unitree-Go1                    | mjlab.tasks.velocity.config.go1.env_cfgs:UNITREE_GO1_FLAT_ENV_CFG                            |
+---+----------------------------------------------------+----------------------------------------------------------------------------------------------+
```

Closes https://github.com/mujocolab/mjlab/issues/301

### DIFF
--- a/src/mjlab/tasks/tracking/config/g1/__init__.py
+++ b/src/mjlab/tasks/tracking/config/g1/__init__.py
@@ -5,8 +5,8 @@ gym.register(
   entry_point="mjlab.envs:ManagerBasedRlEnv",
   disable_env_checker=True,
   kwargs={
-    "env_cfg_entry_point": f"{__name__}.env_cfgs:UNITREE_G1_FLAT_TRACKING_ENV_CFG",
-    "rl_cfg_entry_point": f"{__name__}.rl_cfg:UnitreeG1FlatPPORunnerCfg",
+    "env_cfg_entry_point": f"{__name__}.env_cfgs:G1_FLAT_TRACKING_ENV_CFG",
+    "rl_cfg_entry_point": f"{__name__}.rl_cfg:G1FlatPPORunnerCfg",
   },
 )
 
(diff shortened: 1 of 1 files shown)


brentyi:
Ah, nice catch!

Alternate thought: what if we just remove the entrypoint from the printed table? I'm not a fan of the string-based registration. If we want to print "where" the config was defined/registered one option is we could detect what module the `register()` function was called from and print that (relevant to #300, seems like we've been thinking about similar things).

louislelay:
Oh I see, you already removed it in #300 ahah, I’ll probably close mine in that case! 

AndI don’t really mind if we print the “where”, I mainly use it to quickly access the task IDs.

kevinzakka:
Thanks @louislelay, we're gonna merge #300.

PR #304, opened 2025-11-10
Removed redundant naming from docs

zmk5:
One section of the docs has some redundant naming fixed by your pull request. This will help make things clearer in the docs too.

### DIFF
--- a/src/mjlab/__init__.py
+++ b/src/mjlab/__init__.py
@@ -4,8 +4,6 @@ from pathlib import Path
 
 import warp as wp
 
-__all__ = ["MJLAB_SRC_PATH"]
-
 MJLAB_SRC_PATH: Path = Path(__file__).parent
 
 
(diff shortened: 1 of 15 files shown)


brentyi (approved):
Thanks @zmk5! After discussing with @kevinzakka we decided to go with the `import X as X` convention for exports

zmk5:
Pardon my ignorance on this, but what is the purpose of `import X as X` convection for exports? Just curious to know in case I need to use it on my projects.

brentyi:
This is a "redundant module alias":
- https://typing.python.org/en/latest/spec/distributing.html#import-conventions
- https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface

To summarize, there are two options for designating an imported name as part of a public interface in the typed Python world:
1. Put `"X"` in `__all__ = ["X"]`
2. `import X as X`

Neither are particularly aesthetic to me. But I prefer the latter since it's easier to glance and see which names are vs aren't exported, and if you want to add or remove a name it's all in the same line.

zmk5:
Thank you so much! I really appreciate this!

PR #310, opened 2025-11-13
Fix CartPole documentation and add training tests

FarisHamdi:
### Issue

The CartPole tutorial in `docs/create_new_task.md` resulted in a two errors, one related to joint_names and the other related to the starting position (0, 0, 0), which resulted in a collision throughout training. 

1. **Actuator name didn't match joint name** - XML had `name="slide_velocity"` but the joint was `"slide"`. The code currently needs these to match. 
    -`ValueError: Not all regular expressions are matched! Please check that the regular expressions are correct:
slide_velocity: []
Available strings: ('slide', 'hinge')`

3. **Missing initial state** - Without setting up the initial state the cartpole example would not train successfully. This turned out to be due to the cartpole being stuck below the ground plane

### Fixes

- Fixed the actuator name in the XML example and added a note explaining the need to match actuators to joint names
- Added `init_state` to the `cartpole_constants.py` example  
- Cleaned up imports to match current patterns
- Added a training test that verifies CartPole actually works and hits reward > 3.00

### Testing

Test passes on both GPU and CPU, takes ~42s on CPU. Should work fine in CI.


kevinzakka (changes requested):
Hi @FarisHamdi, thanks for your contribution. Why is there a test file that calls a non-existent cartpole config?

FarisHamdi:
Thank you for the review, and apologies I missed the test assets. The goal here was to add a test to ensure that the cart pole example always trains successfully. Admittedly, this may be a bit too heavy handed. Let me know if you would prefer to not have the test

kevinzakka:
Hi @FarisHamdi, I'm going to make a separate cartpole PR in the future and have the documentation cleverly use that code (I hope through some sphinx / markdown magic). If you don't mind, can we keep this PR purely a documentation fix for now? Appreciate the patience and the really neat test though!

PR #321, opened 2025-11-13
Fix ty issues.

kevinzakka:


src/mjlab/viewer/base.py
```diff
@@ -49,19 +49,18 @@ class EnvProtocol(Protocol):
   `RslRlVecEnvWrapper`, etc."""
 
   device: torch.device
```
brentyi: To avoid the casts above, you can try:

```python
@property
def device(self) -> torch.device | str:
```

The reason the Protocol check fails is that rsl_rl's `VecEnv` annotates the device as `torch.device | str`, and this gets interpreted differently in ty vs pyright. The field needs to be a property to make it read-only => covariant, which makes the protocol compatible with both:
- `device: torch.device` in mjlab's ManagerBasedRlEnv
- `device: torch.device | str` in rsl_rl's `VecEnv`.

kevinzakka: Done.

kevinzakka:
Thanks!

PR #325, opened 2025-11-13
Add learned actuator network model.

kevinzakka:
This PR introduces a new explicit actuator type, `LearnedMlpActuator`, that uses a learned neural network to map joint state history to torques. It’s designed to handle messy dynamics that aren’t well covered by simple analytic models (e.g., nonlinear torque curves, friction and stiction effects, unmodeled compliance, etc.). We’ve bundled a pretrained model from [walk-these-ways](https://github.com/Improbable-AI/walk-these-ways) for use with the Unitree Go1.


kevinzakka:
cc @gmargo11

ax-anoop:
this is insanely cool !

PR #326, opened 2025-11-13
Fix actuator bugs and inconsistencies.

kevinzakka:
- Fix XML actuator doc comment to use joint_name_expr
- Make XML actuators filter to joints with matching XML actuators
- Remove terrain assertion in play script
- Add tests for underactuated XML actuator filtering


Gregwar:
Tested your branch with cartpole example, I confirm that:

* Now works with `joint_names_expr=(r".*",)`
* Can create an env without terrain

PR #334, opened 2025-11-15
Refactor task configs.

kevinzakka:
**TLDR:** This is a structural refactor of how task configs are built. The actual task logic (rewards, observations, actions, etc.) is unchanged.

## Context

We initially used dataclasses with `__post_init__`, but this had fundamental issues:
- `__post_init__` silently overrides constructor arguments
- CLI arguments like `--env.episode-length-s 10.0` get ignored
- Field redeclaration was verbose and error-prone

We refactored to a function-based approach, but it had its own problems:
- Base factory took 10+ robot-specific parameters
- Robot configs had to pass all params upfront (even for defaults)
- Configs were singletons (via `@retval` decorator)
- Creating variants required `deepcopy()` everywhere

## This PR

Simple factory pattern: **base factory creates structure, robot configs mutate and return**.

Base factory defines the task structure:
```python
def make_velocity_env_cfg() -> ManagerBasedRlEnvCfg:
  """Creates base config with placeholder values."""
  # Build observations, actions, rewards, etc.
  return ManagerBasedRlEnvCfg(...)
```

Robot configs customize as needed:
```python
def unitree_go1_rough_env_cfg(play: bool = False) -> ManagerBasedRlEnvCfg:
  cfg = make_velocity_env_cfg()

  # Customize robot-specific parts
  cfg.scene.entities["robot"] = get_go1_robot_cfg()
  cfg.actions["joint_pos"].scale = GO1_ACTION_SCALE
  cfg.rewards["pose"].params["std_standing"] = {...}

  if play:
    cfg.episode_length_s = int(1e9)
    cfg.observations["policy"].enable_corruption = False

  return cfg
```

## What changed

- Base factory takes zero parameters (no more parameter explosion)
- Eliminated `@retval` hack and `deepcopy()` boilerplate
- No more `__post_init__` footguns - mutations are explicit
- Customization is simple: call base factory, mutate what you need, return
- Registry now accepts functions instead of instances (lazy evaluation)
- Play mode is now a `play: bool` parameter instead of duplicate config definitions

## Testing

This refactor also introduces 3 new test files that check that:
- All configs load without errors
- Play/training modes are correctly configured
- Observation corruption enabled/disabled appropriately
- Config structures match between play and training variants

The mutation-based approach is now validated by comprehensive tests, so silent failures get caught.

## Tradeoffs

**What we lose:**
- Mutation-based customization (`cfg.rewards["pose"].params["std_standing"] = {...}`) doesn't give autocomplete or type checking for nested dict keys
- More imperative than declarative (compared to dataclass field definitions)

**What we gain:**
- Mutations are explicit and visible (no `__post_init__` magic)
- Don't need to predict all customization points upfront
- Simpler to understand - no complex inheritance hierarchy to trace
- Tests validate the mutation paths, catching silent failures

Overall, this is a better tradeoff: we give up some type safety at the mutation sites, but gain clarity, flexibility, and comprehensive test coverage. The explicit nature of mutations makes the code easier to understand and modify.

### DIFF
--- a/src/mjlab/scripts/demo.py
+++ b/src/mjlab/scripts/demo.py
@@ -29,7 +29,7 @@ def main() -> None:
       motion_file=motion_path,
       num_envs=8,
       viewer="viser",
-      demo_mode=True,
+      _demo_mode=True,
     ),
   )
   run_play("Mjlab-Tracking-Flat-Unitree-G1-Play", args)
--- a/src/mjlab/scripts/play.py
+++ b/src/mjlab/scripts/play.py
@@ -36,7 +36,9 @@ class PlayConfig:
   video_width: int | None = None
   camera: int | str | None = None
   viewer: Literal["auto", "native", "viser"] = "auto"
-  demo_mode: bool = False
+
+  # Internal flag used by demo script.
+  _demo_mode: tyro.conf.Suppress[bool] = False
 
 
 def run_play(task: str, cfg: PlayConfig):
@@ -57,7 +59,7 @@ def run_play(task: str, cfg: PlayConfig):
     and isinstance(env_cfg.commands["motion"], MotionCommandCfg)
   )
 
-  if is_tracking_task and cfg.demo_mode:
+  if is_tracking_task and cfg._demo_mode:
     # Demo mode: use uniform sampling to see more diversity with num_envs > 1.
     assert env_cfg.commands is not None
     motion_cmd = env_cfg.commands["motion"]


src/mjlab/tasks/registry.py
```diff
@@ -1,16 +1,22 @@
...
+  type | None,
+]
 _REGISTRY: dict[str, EnvRlCfgPair] = {}
 
 
 def register_mjlab_task(
   task_id: str,
-  env_cfg: ManagerBasedRlEnvCfg,
+  env_cfg: ManagerBasedRlEnvCfg | Callable[[], ManagerBasedRlEnvCfg],
   rl_cfg: RslRlOnPolicyRunnerCfg,
```
brentyi, line 20: For consistency, perhaps we should make the RL configs functions as well?

For simplicity, I'm tempted to make it either always expect a factory:
```python
  env_cfg: Callable[[], ManagerBasedRlEnvCfg],
  rl_cfg: Callable[[], RslRlOnPolicyRunnerCfg],
```
or always expect an instance:
```python
  env_cfg: ManagerBasedRlEnvCfg,
  rl_cfg: RslRlOnPolicyRunnerCfg,
```

src/mjlab/tasks/tracking/config/g1/__init__.py
```diff
@@ -2,21 +2,45 @@
...
-  env_cfg=UNITREE_G1_FLAT_TRACKING_ENV_CFG,
+  env_cfg=unitree_g1_flat_tracking_env_cfg,
+  rl_cfg=UNITREE_G1_TRACKING_PPO_RUNNER_CFG,
+  runner_cls=MotionTrackingOnPolicyRunner,
+)
+
+register_mjlab_task(
+  task_id="Mjlab-Tracking-Flat-Unitree-G1-Play",
+  env_cfg=unitree_g1_flat_tracking_env_cfg_play,
+  rl_cfg=UNITREE_G1_TRACKING_PPO_RUNNER_CFG,
```
brentyi: Feels slightly unfortunate that the `Play` tasks also need an RL config

src/mjlab/tasks/velocity/velocity_env_cfg.py
```diff
@@ -252,7 +195,7 @@ def create_velocity_env_cfg(
       weight=1.0,
       params={
         "std": math.sqrt(0.2),
-        "asset_cfg": SceneEntityCfg("robot", body_names=(viewer_body_name,)),
+        "asset_cfg": SceneEntityCfg("robot", body_names=()),  # Set per-robot.
```
brentyi, line 199: I agree aesthetically that large signatures for functions like `create_velocity_env_cfg() ` were unideal, but this "Set per-robot" as a comment does feel more brittle to me... like if the "base" config changes to require more fields that need to be overwritten, it's much easier to have silent failures

src/mjlab/tasks/tracking/config/g1/__init__.py
brentyi: @kevinzakka what do you think about this:
```python
register_mjlab_task(
  task_id="Mjlab-Tracking-Flat-Unitree-G1",
  env_cfg=unitree_g1_flat_tracking_env_cfg(play=False),
  play_env_cfg=unitree_g1_flat_tracking_env_cfg(play=True), # New
  rl_cfg=UNITREE_G1_TRACKING_PPO_RUNNER_CFG,
  runner_cls=MotionTrackingOnPolicyRunner,
)
```

src/mjlab/scripts/demo.py
```diff
@@ -29,10 +29,9 @@ def main() -> None:
       motion_file=motion_path,
       num_envs=8,
       viewer="viser",
-      motion_command_sampling_mode="uniform",
     ),
   )
-  run_play("Mjlab-Tracking-Flat-Unitree-G1", args)
+  run_play("Mjlab-Tracking-Flat-Unitree-G1-Demo", args)
```
brentyi: Since the demo is a one-off thing and the only difference is sampling mode, I still kind of prefer just loading the play config and then overwriting the sampling mode

brentyi (approved):
Looks much better!!

My weak feeling is that it would be nice to drop the `Callable[[], ManagerBasedRlEnvCfg]` from the task registration signature for simplicity; config construction is very fast and forcing all of the factories to run might surface bugs/errors faster.

If every task expects a play variant, perhaps we could just include the play config as part of the task definition?
```python
  env_cfg: ManagerBasedRlEnvCfg,
  play_env_cfg: ManagerBasedRlEnvCfg,
  rl_cfg: RslRlOnPolicyRunnerCfg,
```

kevinzakka:
@brentyi Adopted both suggestions! Removed `Callable` and added `play_env_cfg` parameter to group training/play configs in a single registration. Also discovered and fixed a pre-existing bug where terrain generators were being shared across configs.

src/mjlab/tasks/registry.py
kevinzakka: Made everything an instance!

src/mjlab/tasks/tracking/config/g1/__init__.py
kevinzakka: I like it, and done :)

src/mjlab/scripts/demo.py
kevinzakka: Done.

PR #336, opened 2025-11-16
Simplify multi-GPU training with `torchrunx`

brentyi:
Training on 2 GPUs, before:
```
uv run torchrun \
  --nproc_per_node=2 \
  --no_python \
  train Mjlab-Velocity-Flat-Unitree-G1 \
    --distributed True \
    --env.scene.num-envs 4096
```

Training on 2 GPUs, after:
```
uv run train Mjlab-Velocity-Flat-Unitree-G1 \
    --gpu-ids 0 1 \
    --env.scene.num-envs 4096
```

cc @apoorvkh

### DIFF
--- a/docs/api/distributed_training.md
+++ b/docs/api/distributed_training.md
@@ -41,7 +41,35 @@ uv run train <task-name> \
 - GPU indices are relative to `CUDA_VISIBLE_DEVICES` if set
 - `CUDA_VISIBLE_DEVICES=2,3 uv run train ... --gpu-ids 0 1` uses physical GPUs 2 and 3
 - Each GPU runs the full `num-envs` count (e.g., 2 GPUs × 4096 envs = 8192 total)
-- Single-GPU and CPU modes run directly; multi-GPU uses `torchrunx` for process spawning
+- Single-GPU and CPU modes run directly; multi-GPU uses `torchrunx` for process
+  spawning
+
+## Configuration
+
+**torchrunx Logging:**
+
+By default, torchrunx process logs are saved to `{log_dir}/torchrunx/`. You can
+customize this:
+
+```bash
+# Disable torchrunx file logging.
+uv run train <task-name> --gpu-ids 0 1 --torchrunx-log-dir ""
+
+# Custom log directory.
+uv run train <task-name> --gpu-ids 0 1 --torchrunx-log-dir /path/to/logs
+
+# Or use environment variable (takes precedence over flag).
+TORCHRUNX_LOG_DIR=/tmp/logs uv run train <task-name> --gpu-ids 0 1
+```
+
+The priority is `TORCHRUNX_LOG_DIR` env var, `--torchrunx-log-dir` flag, default
+`{log_dir}/torchrunx`.
+
+**Single-Writer Operations:**
+
+Only rank 0 performs file I/O operations (config files, videos, wandb logging)
+to avoid race conditions. All workers participate in training, but logging
+artifacts are written once by the main process.
 
 ## How It Works
 
--- a/src/mjlab/scripts/train.py
+++ b/src/mjlab/scripts/train.py
@@ -6,12 +6,12 @@ import sys
 from dataclasses import asdict, dataclass, field
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Literal, cast
+from typing import Literal, cast
 
 import tyro
 from rsl_rl.runners import OnPolicyRunner
 
-from mjlab.envs import ManagerBasedRlEnv
+from mjlab.envs import ManagerBasedRlEnv, ManagerBasedRlEnvCfg
 from mjlab.rl import RslRlOnPolicyRunnerCfg, RslRlVecEnvWrapper
 from mjlab.tasks.registry import list_tasks, load_env_cfg, load_rl_cfg, load_runner_cls
 from mjlab.tasks.tracking.mdp import MotionCommandCfg
@@ -23,13 +23,14 @@ from mjlab.utils.wrappers import VideoRecorder
 
 @dataclass(frozen=True)
 class TrainConfig:
-  env: Any
+  env: ManagerBasedRlEnvCfg
   agent: RslRlOnPolicyRunnerCfg
   registry_name: str | None = None
   video: bool = False
   video_length: int = 200
   video_interval: int = 2000
   enable_nan_guard: bool = False
+  torchrunx_log_dir: str | None = None
   gpu_ids: list[int] | Literal["all"] | None = field(default_factory=lambda: [0])
 
   @staticmethod
@@ -40,13 +41,15 @@ class TrainConfig:
     return TrainConfig(env=env_cfg, agent=agent_cfg)
 
 
-def run_train(task_id: str, cfg: TrainConfig) -> None:
+def run_train(task_id: str, cfg: TrainConfig, log_dir: Path) -> None:
   cuda_visible = os.environ.get("CUDA_VISIBLE_DEVICES", "")
   if cuda_visible == "":
     device = "cpu"
     seed = cfg.agent.seed
+    rank = 0
   else:
     local_rank = int(os.environ.get("LOCAL_RANK", "0"))
+    rank = int(os.environ.get("RANK", "0"))
     # Set EGL device to match the CUDA device.
     os.environ["MUJOCO_EGL_DEVICE_ID"] = str(local_rank)
     device = f"cuda:{local_rank}"
@@ -58,7 +61,7 @@ def run_train(task_id: str, cfg: TrainConfig) -> None:
   cfg.agent.seed = seed
   cfg.env.seed = seed
 
-  print(f"[INFO] Training with: device={device}, seed={seed}")
+  print(f"[INFO] Training with: device={device}, seed={seed}, rank={rank}")
 
   registry_name: str | None = None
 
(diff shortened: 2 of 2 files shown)


apoorvkh:
Super cool! I made some suggestions in #338, in case you agree with my thinking!

src/mjlab/scripts/train.py
```diff
@@ -141,6 +142,34 @@ def run_train(task_id: str, cfg: TrainConfig) -> None:
...
+
+  env_vars = {"MUJOCO_GL": "egl"}
+
+  if args.gpus == "all":
+    import torch.cuda
+
+    num_gpus = torch.cuda.device_count()
+  else:
+    num_gpus = len(args.gpus)
+    env_vars["CUDA_VISIBLE_DEVICES"] = ",".join(map(str, args.gpus))
```
brentyi: One concern I have here is that folks might be surprised to find the `CUDA_VISIBLE_DEVICES` variable in their env ignored:
```
export CUDA_VISIBLE_DEVICES=0
uv run train [...]
```
This is one reason why I had `--num-gpus` originally, it seems more compatible with "standard" PyTorch workflows?

kevinzakka (approved):

src/mjlab/scripts/train.py
```diff
@@ -141,6 +142,34 @@ def run_train(task_id: str, cfg: TrainConfig) -> None:
...
+
+  print(f"[INFO] Launching training with {num_gpus} GPUs", flush=True)
+  launcher = torchrunx.Launcher(
```
brentyi: I'm lightly biased toward only using `torchrunx` when doing >1 GPU because of:
- Startup overhead
- Simplicity / reducing indirection
    - Confidence that things like `breakpoint()`, `ipdb.set_trace()`, will just work

kevinzakka: +1

src/mjlab/scripts/train.py
kevinzakka: Agreed as well, CUDA_VISIBLE_DEVICES has always been my safety net for running things on multi-GPU machines and I'd expect it to behave the same in mjlab.

apoorvkh: Good point! At your discretion, here are a few viable options. I prefer (2) for its flexibility.

(1) Making `CUDA_VISIBLE_DEVICES` the only control.

(2) Use `gpus: list[int] | "all"` to pass indices corresponding to `CUDA_VISIBLE_DEVICES`.

(3) Use `num_gpus: int` to set indices `[0, ..., num_gpus - 1]` corresponding to `CUDA_VISIBLE_DEVICES`.

src/mjlab/scripts/train.py
apoorvkh: Agree with your points (especially for debugging)! Will add debugging support in `torchrunx` eventually. I do think you get a nice "process-level reliability" via `torchrunx`, perhaps at a tradeoff with overhead. But feel free to revert this for now :)

kevinzakka:
@brentyi @apoorvkh I implemented both pieces of feedback from the review.

1. The `--gpus` parameter now specifies indices into the existing `CUDA_VISIBLE_DEVICES` environment variable rather than overriding it. For example, `CUDA_VISIBLE_DEVICES=1,3 uv run train ... --gpus 0 1` will use physical gpus 1 and 3.
2. Single-gpu training now calls `run_train_on_gpu()` directly without `torchrunx` overhead, preserving debugging tool support (e.g., `ipdb`).

Note the gpu selection logic has been extracted to `mjlab.utils.gpu.select_gpus()` with some corresponding tests in `test_gpu_selection.py`.

src/mjlab/scripts/train.py
kevinzakka: Thanks @apoorvkh, will happily switch to `torchrunx` for the single-GPU case when debugging is properly supported!

src/mjlab/scripts/train.py
kevinzakka: I also prefer 2 and went ahead and implemented it. Thanks @apoorvkh!

kevinzakka (approved):

src/mjlab/scripts/train.py
brentyi: What would be the new command to run for training on CPU?

kevinzakka: Training on CPU is not encouraged but, we could support:

1) CUDA_VISIBLE_DEVICES="" uv run train ..., or
2) uv run train Mjlab-Tracking-Flat-Unitree-G1 --gpus cpu ...

Thoughts?

kevinzakka: I just tried training on CPU and it's ridiculously slow and the kernels keep reloading (which points to a potential warp issue). So I upgrade my "not encouraged" to "strongly discouraged".

brentyi: Yeah I agree we wouldn't want to do it in practice, it's just nice to make the CLI interface more general in case we want to either dry run to test changes locally or if one day mjwarp is faster on CPU!

For the CLI interface itself, maybe:
- `--devices 0 1`, `--devices cpu` ?
- `--gpus 0 1`, `--gpus None` ?
- `--devices cuda:0 cuda:1`, `--devices cpu`?

brentyi: Other suggestion: `--gpu-ids` instead of `--gpus` to make it clear it's not a count?

src/mjlab/scripts/train.py
```diff
@@ -141,6 +143,39 @@ def run_train(task_id: str, cfg: TrainConfig) -> None:
...
+
+    print(f"[INFO] Launching training with {num_gpus} GPUs", flush=True)
+    launcher = torchrunx.Launcher(
+      hostnames=["localhost"],
+      workers_per_host=num_gpus,
+      backend=None,  # Let rsl_rl handle process group initialization.
+      extra_env_vars=env_vars,
+    )
+    launcher.run(run_train_on_gpu, task_id, args)
+
```
apoorvkh, line 205: Could also write this as:

```python
  os.environ["CUDA_VISIBLE_DEVICES"] = ",".join(map(str, selected_gpus))
  os.environ["MUJOCO_GL"] = "egl"

  if num_gpus == 1:
    run_train_on_gpu(task_id, args)
  else:
    # Multi-GPU: use torchrunx.
    import torchrunx

    # torchrunx redirects stdout to logging.
    logging.basicConfig(level=logging.INFO)

    print(f"[INFO] Launching training with {num_gpus} GPUs", flush=True)
    launcher = torchrunx.Launcher(
      hostnames=["localhost"],
      workers_per_host=num_gpus,
      backend=None,  # Let rsl_rl handle process group initialization.
      copy_env_vars=torchrunx.DEFAULT_ENV_VARS_FOR_COPY + ("MUJOCO*",)
    )
    launcher.run(run_train_on_gpu, task_id, args)
```

Since `torchrunx` propagates a subset of current environment variables ([common to Python/PyTorch](https://torchrun.xyz/api.html#torchrunx.Launcher.copy_env_vars)) and we can tell it to also copy `MUJOCO` env vars.

apoorvkh: Also `torchrunx.Launcher(...).run(...)` if you prefer

kevinzakka: Done!

src/mjlab/scripts/train.py
kevinzakka: Done!

src/mjlab/scripts/train.py
kevinzakka: Thanks again @apoorvkh!

apoorvkh: No problem! Last thing is that you'll notice that `torchrunx` saves files to `./torchrunx_logs/` by default. You can change this path via `$TORCHRUNX_LOG_DIR`. I am about to push an update where `TORCHRUNX_LOG_DIR=""` results in no file logging.

docs/api/distributed_training.md
```diff
@@ -1,43 +1,57 @@
 # Distributed Training
 
-mjlab supports multi-GPU distributed training using `torchrun`. Distributed
+mjlab supports multi-GPU distributed training using `torchrunx`. Distributed
```
apoorvkh: Would you kindly add a link to torchrunx here? Thanks! You could use: https://github.com/apoorvkh/torchrunx

docs/faq.md
```diff
@@ -40,7 +40,8 @@ On par or faster based on our experience over the last few months.
 
 ### Does mjlab support multi-GPU training?
 
-Yes. mjlab supports multi-GPU distributed training with `torchrun`. See the
+Yes. mjlab supports multi-GPU distributed training with `torchrunx`. Use
```
apoorvkh: And maybe also link here?

src/mjlab/scripts/train.py
apoorvkh: Ability to disable file logging via `TORCHRUNX_LOG_DIR=""` just released in `torchrunx==0.3.4`.

docs/faq.md
kevinzakka: Done!

docs/api/distributed_training.md
kevinzakka: Of course, done!

src/mjlab/scripts/train.py
kevinzakka: Cool, I've added `--torchrunx-log-dir` flag (defaults to `{log_dir}/torchrunx`) and it respects `TORCHRUNX_LOG_DIR` env var so that means the user can disable `torchrunx` logging either via the env var or via `--torchrunx-log-dir ""`

apoorvkh: Great! And if you want more extreme control, please see: https://torchrun.xyz/usage/logging.html

If you tell me exactly where you want the logs to go, I can help write custom handlers for that!

PR #338, opened 2025-11-16
Proposed changes for `torchrunx`-based training script

apoorvkh:
Some definitely opinionated suggestions! Please feel free to consider as you see fit.

- I've removed the `device` and `num_gpus` arguments in favor of a single `gpus` argument (referring to GPU IDs)
- I've baked the environment variables into `torchrunx.Launcher`
- I separated the logic for the CLI and `torchrunx` into two functions. This would permit a downstream user to invoke training in their own script (e.g. as part of a more complicated workflow) via `launch_training(task_id, args)`.
-  I think you can use `torchrunx` even in the single-GPU case.
    - This may be desirable because it wraps the underlying training runtime in a process. If a user wants to call `launch_training` as above, this would isolate potential memory leaks to that wrapped process and also make both [system failures](https://torchrun.xyz/api.html#torchrunx.WorkerFailedError) and regular exceptions (e.g. CUDA OOM) catchable. This permits users to do more things in the same script (e.g. multiple training runs in succession).

### DIFF
--- a/src/mjlab/scripts/train.py
+++ b/src/mjlab/scripts/train.py
@@ -3,10 +3,10 @@
 import logging
 import os
 import sys
-from dataclasses import asdict, dataclass
+from dataclasses import asdict, dataclass, field
 from datetime import datetime
 from pathlib import Path
-from typing import Annotated, Any, cast
+from typing import Any, Literal, cast
 
 import tyro
 from rsl_rl.runners import OnPolicyRunner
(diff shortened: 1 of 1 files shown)


src/mjlab/scripts/train.py
```diff
@@ -147,6 +142,34 @@ def run_train(task_id: str, cfg: TrainConfig) -> None:
...
+
+  # torchrunx redirects stdout to logging.
+  logging.basicConfig(level=logging.INFO)
+
+  env_vars = {"MUJOCO_GL": "egl"}
+
+  if args.gpus == "all":
+    import torch.cuda
+
+    num_gpus = torch.cuda.device_count()
```
apoorvkh, line 158: Only counting GPUs to print a specific `num_gpus` below. If you are happy to just say `"Launching training with all GPUs"`, you can set `workers_per_host="gpu"` when `args.gpus == "all"`.

brentyi:
+1 looks great! Let's merge this and then discuss in the other PR 🙂

PR #354, opened 2025-11-21
Fix position actuator control clipping bug.

kevinzakka:
Position actuators with `inheritrange=1` were getting `ctrllimited=True` when `autolimits=True`, causing control setpoints to be clipped to joint limits internally. This broke the assumption that policies can output setpoints outside the kinematic limits of the joint.

We now set `ctrllimited=False` for position actuators and remove the inheritrange parameter. Tests verify unclipped setpoints generate 3x more torque and all G1 actuators are configured correctly.


kevinzakka:
Extremely ashamed of this bug 🫠

louislelay:
Don't be, bravo for catching it !!!

PR #355, opened 2025-11-21
Fixes ONNX export crash for robots with non-actuated joints

louislelay:
Bug introduced by https://github.com/mujocolab/mjlab/pull/352. ONNX export was failing with a `KeyError` in `get_base_metadata` when `robot.joint_names` contained joints without corresponding actuators. This PR makes `ctrl_ids_natural` robust by skipping joints that are not present in `joint_name_to_ctrl_id`, so models with non-actuated joints can be exported without errors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,11 +27,11 @@ jobs:
         uses: astral-sh/setup-uv@v6
         with:
           enable-cache: true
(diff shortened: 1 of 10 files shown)


louislelay:
Kangaroo robot really is a good edge case @kevinzakka! :laughing:

PR #360, opened 2025-11-24
Clarify and improve entity positioning.

kevinzakka:


src/mjlab/viewer/viser_conversions.py
```diff
@@ -434,10 +434,29 @@ def rotation_matrix_from_vectors(
...
+    return True
+
+  # Welded to world AND not mocap = truly static
+  is_welded = mj_model.body_weldid[body_id] == 0
+  is_mocap = mj_model.body_mocapid[body_id] >= 0
+
+  return is_welded and not is_mocap
+
+
 def is_fixed_body(mj_model: mujoco.MjModel, body_id: int) -> bool:
```
louislelay: Nit: maybe rename this to is_welded_to_world? Here we only check body_weldid == 0 (regardless of mocap), while above “fixed” is defined as welded and not a mocap body. The new name would make the distinction clearer.

kevinzakka: Thanks, I've improved both function names. Should be clearer now I hope.

PR #365, opened 2025-11-25
Fix observation delay buffer double-push bug.

### LINKED ISSUE #364: Observation delays
Hello,

Below are several related problems in the way delays are handled

## Observations are pushed twice in delay buffer

In `manager_based_rl_env`, there ate two calls to `observation_manager.compute` (in `reset` and `step`)

But there is a side effect here in `observation_manager.compute`:
https://github.com/mujocolab/mjlab/blob/be706226f04d2e40d83879e153c8120a56aef80f/src/mjlab/managers/observation_manager.py#L173

This cause each observation to be pushed twice in the delay buffer. This will somehow divide by two the actual delay obtained

## Consistency with delayed actuators

The delayed actuators have delays expressed in physics timesteps and the observation delays are expressed in control steps. More than a consistency issue, this is also a limitation in the granularity of delays we can simulate when using decimation factors

kevinzakka:
Return cached observations when compute(update_history=False) is called, preventing duplicate pushes to delay buffer when get_observations() is called after step().

Fixes #364.

### DIFF
--- a/src/mjlab/managers/observation_manager.py
+++ b/src/mjlab/managers/observation_manager.py
@@ -122,6 +122,9 @@ class ObservationManager(ManagerBase):
   # Methods.
 
   def reset(self, env_ids: torch.Tensor | slice | None = None) -> dict[str, float]:
+    # Invalidate cache since reset envs will have different observations.
+    self._obs_buffer = None
+
     for group_name, group_cfg in self._group_obs_class_term_cfgs.items():
       for term_cfg in group_cfg:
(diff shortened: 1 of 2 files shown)


Gregwar:
Hello,

The `update_history` flag is passed to `True` by the env in both reset and step, shouldn't it be `False` in reset for this fix to work ?

https://github.com/mujocolab/mjlab/blob/2d944bca9c15b36aa78e9d92979e612f2fd3ff3b/src/mjlab/envs/manager_based_rl_env.py#L270

PR #368, opened 2025-11-26
Add EntityCfg.build() method for custom entity subclasses.

### LINKED ISSUE #367: Inheriting Entity class
In IsaacLab, the `ArticulationCfg` [specifies the target class](https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab/isaaclab/assets/articulation/articulation_cfg.py#L39), which we can modify to inform the scene builder which subclass we want to instantiate the asset into.

For example, I can create a custom Articulation wrapper by doing:

```python
class MyCustomArticulation(Articulation):
    def __init__(self, cfg: "MyCustomArticulationCfg"):
        super().__init__(cfg)

        # add type hinting
        self.cfg: MyCustomArticulationCfg


@configclass
class MyCustomArticulationCfg(ArticulationCfg):
    """Configuration parameters for an articulation."""

    class_type: type = MyCustomArticulation

```

I don't think MjLab's `Entity` class has the similar field. So I am wondering how to achieve such functionality?

kevinzakka:
Allows users to create custom Entity subclasses with domain-specific properties and config fields.

Closes #367.

### DIFF
--- a/src/mjlab/scene/scene.py
+++ b/src/mjlab/scene/scene.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any
+from typing import Any, Callable
 
 import mujoco
 import mujoco_warp as mjwarp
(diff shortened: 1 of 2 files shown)


T-K-233 (approved):
LGTM.

PR #371, opened 2025-11-27
Fixes ObservationManager mutating shared term configs between groups

louislelay:
When observation groups share the same term configs (e.g., the critic reuses the policy’s terms), the `ObservationManager` mutates those shared config objects in place. This causes unintended side effects across groups.

Concretely, with something like:

```python
policy_terms = {
    "...": ...,
}
critic_terms = {
    **policy_terms,
    "...": ...,
}

observations = {
    "policy": ObservationGroupCfg(
        terms=policy_terms,
        history_length=5,
        concatenate_terms=True,
        enable_corruption=True,
    ),
    "critic": ObservationGroupCfg(
        terms=critic_terms,
        concatenate_terms=True,
        enable_corruption=False,
    ),
}
```

`ObservationManager` processes the `"policy"` group first and sets `term_cfg.history_length = 5` on the *shared* term objects.
When it later processes the `"critic"` group, those same objects already have `history_length=5`, so the critic incorrectly gets a history applied as well.

This leads to a mismatch between training and evaluation:

**Training (correct):**

```text
Active Observation Terms in Group: 'critic' (shape: (111,))
```

**Play / evaluation (broken):**

```text
Active Observation Terms in Group: 'critic' (shape: (507,))
```

Here, `507 = 111 * history_length=5`, i.e., the critic’s observations are unintentionally expanded.

This then surfaces as a shape mismatch when loading the checkpoint:

```text
RuntimeError: Error(s) in loading state_dict for ActorCritic:
  size mismatch for critic.0.weight: copying a param with shape torch.Size([512, 111]) from checkpoint, the shape in current model is torch.Size([512, 507]).
  size mismatch for critic_obs_normalizer._mean: copying a param with shape torch.Size([1, 111]) from checkpoint, the shape in current model is torch.Size([1, 507]).
  size mismatch for critic_obs_normalizer._var: copying a param with shape torch.Size([1, 111]) from checkpoint, the shape in current model is torch.Size([1, 507]).
  size mismatch for critic_obs_normalizer._std: copying a param with shape torch.Size([1, 111]) from checkpoint, the shape in current model is torch.Size([1, 507]).
```

### Fix

Ensure that each group operates on its own copy of the term config instead of mutating the shared object:

```python
for term_name, term_cfg in group_cfg.terms.items():
    if term_cfg is None:
        continue

    # Copy to avoid mutating the original config
    term_cfg = copy(term_cfg)

    # Mutations now only affect this group's copy
    if group_cfg.history_length is not None:
        term_cfg.history_length = group_cfg.history_length
```

By copying the term configs per group, we prevent cross-group contamination and keep policy and critic observation shapes consistent between training and evaluation.

### DIFF
--- a/src/mjlab/managers/observation_manager.py
+++ b/src/mjlab/managers/observation_manager.py
@@ -236,6 +236,7 @@ class ObservationManager(ManagerBase):
           continue
 
         from copy import copy
+
         term_cfg = copy(term_cfg)
 
         self._resolve_common_term_cfg(term_name, term_cfg)


kevinzakka:
See #373.

PR #373, opened 2025-11-28
Fix config mutation in managers by deepcopying cfg on init.

kevinzakka:
- `deepcopy` cfg dict in all managers to prevent input mutation
- `deepcopy` term configs in ObservationManager to prevent cross-group contamination
- Add comprehensive immutability tests
- Rename `task`-->`task_id` in `play.py` for consistency

### DIFF
--- a/src/mjlab/managers/curriculum_manager.py
+++ b/src/mjlab/managers/curriculum_manager.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from copy import deepcopy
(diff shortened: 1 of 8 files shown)


louislelay (commented):
Biased ofc but lgtm.

brentyi:
This is fine with me, but is it not possible to avoid the mutation to begin with? It seems like a code smell.

kevinzakka:
Let's revisit before stable release! Going to merge for now.

PR #374, opened 2025-11-30
chore: add .DS_Store to .gitignore

Nagi-ovo:
Added .DS_Store to .gitignore to avoid committing macOS system files during local development.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ dependencies = [
     "viser>=1.0.16",
     "moviepy",
     "tensordict",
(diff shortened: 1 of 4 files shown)


kevinzakka:
Hi @Nagi-ovo, thanks for the PR! I'd prefer to keep the .gitignore project specific and not OS-specific. Users should put that in their global gitignore or more carefully add files when creating commits.

Nagi-ovo:
That makes sense — I’ll switch to using a global gitignore as well :)

PR #375, opened 2025-11-30
Add manipulation task.

kevinzakka:
This PR adds the [YAM robot](https://github.com/google-deepmind/mujoco_menagerie/tree/main/i2rt_yam) to the asset zoo and introduces a cube lifting example as a pedagogical reference for setting up manipulation tasks in mjlab.

https://github.com/user-attachments/assets/b2e2e9fd-15f5-4a07-b38f-5b77d028535d

The main infrastructure addition is automatic mocap wrapping for fixed-base entities. Previously, fixed-base robots couldn't be positioned per-environment since they're welded to world. Now, when you load a fixed-base robot, it automatically gets wrapped in a mocap body so each env can have its robot at a different position. The reset logic handles this transparently - you just set `init_state` as usual and it works.

The lift cube task uses a simple reward: a staged term that first rewards getting close to the cube, then lifting it toward the goal. Regularization comes from joint limit and action rate penalties, plus a joint velocity penalty that ramps up over training to encourage slower, more graceful movements.

**Note:** Due to a limitation of the native MuJoCo viewer, the robot base won't render correctly when visualizing more than one parallel environment. The viser viewer does not suffer from this issue, as shown below.

https://github.com/user-attachments/assets/85c07a3b-3fd6-4670-923f-76d504f4f418


louislelay:
👏👏

cmcgartoll:
@kevinzakka  This is great, I am exploring better sim options for manipulation for my lab and am interested in mjlab. I know that manipulation is in beta in this repo, so would be interested to help push it forward. Are there any active manipulation-related projects for mjlab currently? Would love to add some manipulation assets such as, xarm, ability hand, xhand, etc., to start.

louislelay:
Hello @cmcgartoll, last time this was discussed, the idea was to keep the repository as slim as possible to make installation easier. So we do not plan to support many robots by default, but we are very interested in contributions via external repositories and in seeing how you use mjlab! If you implement this in an external repository, please feel free to share it in the discussions :)

cmcgartoll:
Hi @louislelay, I see, makes sense! Will give things a try

Issue #378, opened 2025-12-01
Google Colab and wandb prompt error


ivrolan:
Hi,

I tried first running the Colab and I got the following error:

```bash
wandb.errors.errors.UsageError: api_key not configured (no-tty). call wandb.login(key=[your_api_key])
```

<details>
<summary>Toggle output cell </summary>

```
Installed 121 packages in 1.08s
🎮 Setting up MJLab demo with pretrained tracking policy...
📥 Downloading demo_ckpt.pt...

demo_ckpt.pt:   0%|          | 0.00/7.39M [00:00<?, ?B/s]
demo_ckpt.pt:  54%|█████▍    | 4.01M/7.39M [00:00<00:00, 35.1MB/s]
demo_ckpt.pt: 100%|██████████| 7.39M/7.39M [00:00<00:00, 57.7MB/s]
✅ demo_ckpt.pt cached at /tmp/mjlab_cache/demo_ckpt.pt
📥 Downloading demo_motion.npz...

demo_motion.npz:   0%|          | 0.00/11.2M [00:00<?, ?B/s]
demo_motion.npz:  63%|██████▎   | 7.11M/11.2M [00:00<00:00, 74.5MB/s]
demo_motion.npz: 100%|██████████| 11.2M/11.2M [00:00<00:00, 102MB/s] 
✅ demo_motion.npz cached at /tmp/mjlab_cache/demo_motion.npz
[INFO]: Using device: cuda:0
[INFO]: Parsing configuration from: mjlab.tasks.tracking.config.g1.flat_env_cfg:G1FlatEnvCfg_PLAY
[INFO]: Parsing configuration from: mjlab.tasks.tracking.config.g1.rl_cfg:G1FlatPPORunnerCfg
[INFO]: Loading experiment from: /content/logs/rsl_rl/g1_tracking
[INFO]: Loading checkpoint: /tmp/mjlab_cache/demo_ckpt.pt
[INFO]: Using motion file from CLI: /tmp/mjlab_cache/demo_motion.npz
No seed set for the environment.
[INFO]: Scene manager: <mjlab.scene.scene.Scene object at 0x7a7e4a0ae7b0>
Warp 1.11.0.dev20251130 initialized:
   Git commit: af1e0dd7ee79bb6699d81cd0cdb060ac5c019e32
   CUDA Toolkit 12.9, Driver 12.4
   Devices:
     "cpu"      : "x86_64"
[...]

PR #381, opened 2025-12-01
Fix demo notebook

ivrolan:
Add cells to use wandb offline or to login before running the demo.

Solves #378 . To keep it general, I added the two possible solutions:

1. Running `wandb offline` and only storing data locally, which does not need login
2. Running `wandb login` to prompt for an API key before running the demo.

It seemed that the login prompt was not caught in the cell, as the stdout etc is being read to know when to stop executing the cells.

 Additionally, I added some markdown cells to give more context. If you find it confusing we can also choose only one option to keep! I think `wandb offline` is the one that adds the least friction for a new user.

### DIFF
--- a/notebooks/demo.ipynb
+++ b/notebooks/demo.ipynb
@@ -1,6 +1,45 @@
 {
  "cells": [
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Use W&B offline"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!wandb offline"
+   ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


kevinzakka (approved):
Thanks @ivrolan!

ivrolan:
Just forced push to remove minor changes when opening the notebook with my environment in VSCode. Now it's cleaner!

Old diff contained these changes that were not needed:

```
@@ -77,11 +116,13 @@
    "provenance": []
   },
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "base",
+   "language": "python",
    "name": "python3"
   },
   "language_info": {
-   "name": "python"
+   "name": "python",
+   "version": "3.13.2"
   }
  },

```

kevinzakka:
Thanks again @ivrolan!

PR #385, opened 2025-12-02
Minor improvements in the Entity and action manager and circular buffer

mktk1117:
- Modified the Entity to add few properties and write_ctrl_to_sim function for flexibility.

- Added prev2_action in the action manager to be able to calculate action acceleration for penalty.

- For more flexible implementation which could update ctrl_ids etc when not using Actuator classes and directly controlling mujoco actuators.
```
@dataclass(frozen=False)
class EntityIndexing:
```

- Removed unnecessary clearing of the buffer in the reset for more throughput.

### DIFF
--- a/src/mjlab/entity/entity.py
+++ b/src/mjlab/entity/entity.py
@@ -243,8 +243,8 @@ class Entity:
     return tuple(g.name.split("/")[-1] for g in self.spec.geoms)
 
   @property
-  def tendon_names(self) -> list[str]:
-    return [t.name.split("/")[-1] for t in self._spec.tendons]
+  def tendon_names(self) -> tuple[str, ...]:
+    return tuple(t.name.split("/")[-1] for t in self._spec.tendons)
 
   @property
   def site_names(self) -> tuple[str, ...]:
@@ -504,7 +504,7 @@ class Entity:
 
   def write_ctrl_to_sim(
     self, ctrl: torch.Tensor, ctrl_ids: torch.Tensor | slice | None = None
-    ) -> None:
+  ) -> None:
     """Write control inputs to the simulation.
 
     Args:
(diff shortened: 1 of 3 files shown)


kevinzakka (commented):
Hi @mktk1117, thank you for the contribution! I have some high-level comments and questions:

1. `prev2_action` naming: Can we consider something clearer like prev_prev_action? Also, are you missing an `action_acc_l2` reward function that uses this? Something like: `act - 2 * prev_action + prev2_action`?
2. `joint_vel_l2` docstring: says "joint accelerations" but computes velocities.
3. tendon_names without find_tendons: Every other *_names property has a corresponding find_* method. Should add find_tendons for consistency.
4. write_ctrl_to_sim: Missing type hints for ctrl and ctrl_ids.
5. EntityIndexing frozen=False: This makes previously immutable indexing state mutable. What's the specific use case that requires mutation after construction?
6. Circular buffer reset: This change is not safe. Two existing callers read after reset without appending first:
    - DelayBuffer.compute() reads without appending. Its docstring explicitly says "compute() returns zeros for reset rows".
    - ObservationManager.compute_group() skips append when update_history=False, but buffer is still "initialized" after reset.

mktk1117:
Thanks for the comments!!
1. I considered both and chose prev2_action for shorter name. I renamed to prev_prev_action
2. Fixed this, thanks!
3. I added this function.
4. Added type hints.
5. Currently, all ctrl_ids etc, is only set when is_actuated is True. And this is_actuated is only True if the actuators are set through cfg and Actuator class. Also soft_joint_pos_limit is only set to _non_free_joints, etc. All of these assume that actuator is directly actuating joints, which becomes a issue with tendon actuators for example.  frozen=False can add flexibility to create a custom action class which can overwrite and directly control the actuators defined by mujoco without going through the mjlab api. If users want something complicated which is not supported by the api, this can give users flexibility without the need to modify the mjlab codebase itself. 
6. Good point. I think it's a design decision, but isn't it anyway dangerous to return zero values? If there's no buffer, it shouldn't return 0, but give the latest value. Otherwise, the observation becomes all zeros, which can also affect the training.

mktk1117:
For 6. I'll put back for now, and probably open a new PR if needed for performance.

kevinzakka:
Closing in favor of 391 (I screwed up the push).

PR #389, opened 2025-12-02
Refactors LiftingCommand to PositionCommand

louislelay:
`LiftingCommand` was handling position targets, cube randomization, and success metrics. This PR separates those concerns: the command (renamed `PositionCommand`) now only provides position targets, cube randomization moves to an interval event, and success metrics move to a class-based `bring_object_reward`. The cube also gets an explicit init state in its entity config.


kevinzakka:
Minor comment: It's technically not a pose command because we are only specifying a position for the cube.

louislelay:
You're completely right!

PR #392, opened 2025-12-03
Add time_major flag for observation history concatenation (matching leggedLab) 

ax-anoop:
For context, time major is used in legged lab by default, hence the confusion with sim2sim: https://github.com/Hellod035/LeggedLab/blob/main/legged_lab/envs/base/base_env.py

Add a `time_major` option to ObservationGroupCfg that controls how observation history is concatenated when `concatenate_terms=True`.

When `time_major=False` (default), uses term-major ordering:
  [A_t0, A_t1, ..., A_tH-1, B_t0, B_t1, ..., B_tH-1, ...]

When `time_major=True`, uses time-major ordering:
  [A_t0, B_t0, ..., A_t1, B_t1, ..., A_tH-1, B_tH-1, ...]

This is useful when temporal locality across terms is preferred over grouping all history for each term together.

### DIFF
--- a/src/mjlab/managers/manager_term_config.py
+++ b/src/mjlab/managers/manager_term_config.py
@@ -134,6 +134,9 @@ class ObservationGroupCfg:
   enable_corruption: bool = False
   history_length: int | None = None
   flatten_history_dim: bool = True
+  time_major: bool = False
+  """When True with history, concatenates as [A_t0, B_t0, ..., A_t1, B_t1, ...] (time-major).
+  When False (default), concatenates as [A_t0, A_t1, ..., B_t0, B_t1, ...] (term-major)."""
 
 
 ##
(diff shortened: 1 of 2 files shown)


kevinzakka:
Hi @ax-anoop, I want to err on the side of minimal in mjlab. Can you convince me there is a need to support an extra concatenation style (not just to match another library)?

ax-anoop:
https://github.com/isaac-sim/IsaacLab/issues/1208

I would say its totally worth it, espesially since alot of publically trained policies rely of leggedlab, just gives more options for sim2sim :) 

In our experience & we've trained alot of policies on alot of robots now, this is a nice / useful feature. There is also the aspect of readibility when analyzing observations for trouble shooting, we personally prefer time major for that reason. 

No worries if you don't think its needed.

kevinzakka:
Hi @ax-anoop,

While I appreciate the effort, I'm going to decline this PR in favor of keeping mjlab minimal. One of our core goals when creating mjlab was to avoid the configuration sprawl that makes codebases like Isaac Lab difficult to maintain. Every additional flag we add is a maintenance burden and increases the surface area for bugs.

For cases like sim2sim transfers where you need a specific observation layout, I'd suggest reshaping the observations in your policy wrapper/training code. This keeps mjlab focused on the common case while giving you full flexibility in your own code.

Thanks again for understanding!

PR #410, opened 2025-12-12
Merge per-entity keyframes into single scene keyframe.

### LINKED ISSUE #263: Fallback to model keyframe when initial joint positions are not specified in the robot config
Similarly to how we default to the model’s actuators when none are specified in the robot config, we should also fall back to the model’s keyframe for initial joint positions. Manually copying joint poses is usually quick, but on robots with many joints it becomes tedious and error-prone.

kevinzakka:
Previously, each entity created its own keyframe, resulting in multiple separate keyframes when a scene had multiple entities. Each keyframe only contained that entity's qpos/ctrl, with other entities at default values.

Now, the Scene extracts and merges all entity keyframes into a single `init_state` keyframe containing the combined initial state for all entities.

Also adds support for `joint_pos=None` in `EntityCfg.InitialStateCfg` to use the model's existing keyframe instead of manually specifying joint positions. Errors if the model has no keyframe.

Fixes #263.

### DIFF
--- a/src/mjlab/scene/scene.py
+++ b/src/mjlab/scene/scene.py
@@ -1,9 +1,11 @@
+import warnings
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Callable
 
 import mujoco
 import mujoco_warp as mjwarp
+import numpy as np
 import torch
 
 from mjlab.entity import Entity, EntityCfg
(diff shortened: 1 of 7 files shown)


src/mjlab/scene/scene.py
```diff
@@ -150,11 +151,25 @@ def write_data_to_sim(self) -> None:
...
+    # Order matters: qpos/ctrl are concatenated in entity iteration order.
+    key_qpos: list[np.ndarray] = []
+    key_ctrl: list[np.ndarray] = []
     for ent_name, ent_cfg in self._cfg.entities.items():
       ent = ent_cfg.build()
       self._entities[ent_name] = ent
+      # Extract keyframe before attach (must delete before attach to avoid corruption).
+      if ent.spec.keys:
+        key_qpos.append(np.array(ent.spec.keys[0].qpos))
+        key_ctrl.append(np.array(ent.spec.keys[0].ctrl))
```
louislelay, line 171: If the model contains multiple keyframes, it always enforces the first one. We should document this.

PR #416, opened 2025-12-15
Support DelayedActuator in randomize_pd_gains

ax-anoop:
Unwrap DelayedActuator to access the base actuator before applying PD gain randomization. This allows randomize_pd_gains to work with actuators wrapped in DelayedActuatorCfg.

Not sure if this is the best way to fix, but works for me. 

issue: 

Screenshot 2025-12-14 at 7 28 51 PM

### DIFF
--- a/src/mjlab/envs/mdp/events.py
+++ b/src/mjlab/envs/mdp/events.py
@@ -517,6 +517,8 @@ def randomize_pd_gains(
 
   if isinstance(asset_cfg.actuator_ids, list):
     actuators = [asset.actuators[i] for i in asset_cfg.actuator_ids]
+  elif isinstance(asset_cfg.actuator_ids, slice):
+    actuators = asset.actuators[asset_cfg.actuator_ids]
   else:
     actuators = [asset.actuators[asset_cfg.actuator_ids]]
 


src/mjlab/envs/mdp/events.py
```diff
@@ -517,7 +518,10 @@ def randomize_pd_gains(
   if isinstance(asset_cfg.actuator_ids, list):
     actuators = [asset.actuators[i] for i in asset_cfg.actuator_ids]
   else:
-    actuators = asset.actuators[asset_cfg.actuator_ids]
+    actuators = [asset.actuators[asset_cfg.actuator_ids]]
+
+  # Unwrap DelayedActuators to access base actuators.
+  actuators = [a._base_actuator if isinstance(a, DelayedActuator) else a for a in actuators]
```
kevinzakka: Can you add base_actuator as a property and use it here instead of accessing the private attribute?

kevinzakka:
Hi @ax-anoop, the linter-formatter test failed. Can you follow the instructions [here](https://github.com/mujocolab/mjlab?tab=readme-ov-file#development)?

ax-anoop:
sorry is there still an issue ? I'm a bit new to making PRs on public projects

src/mjlab/envs/mdp/events.py
```diff
@@ -517,7 +518,12 @@ def randomize_pd_gains(
   if isinstance(asset_cfg.actuator_ids, list):
     actuators = [asset.actuators[i] for i in asset_cfg.actuator_ids]
   else:
-    actuators = asset.actuators[asset_cfg.actuator_ids]
+    actuators = [asset.actuators[asset_cfg.actuator_ids]]
```
louislelay, line 523: Hello @ax-anoop, the test is failing because when actuator_ids is a slice, asset.actuators[slice(...)] already returns a list. Wrapping that in [...] creates a nested list, which breaks the subsequent loop.

You need to handle all three cases:

```
if isinstance(asset_cfg.actuator_ids, list):
    actuators = [asset.actuators[i] for i in asset_cfg.actuator_ids]
elif isinstance(asset_cfg.actuator_ids, slice):
    actuators = asset.actuators[asset_cfg.actuator_ids] 
else:
    actuators = [asset.actuators[asset_cfg.actuator_ids]] 
```

PR #424, opened 2025-12-18
Fix encoder bias shifting joint limits incorrectly.

### LINKED ISSUE #422: Encoder Bias Logic Incorrectly Shifts Physical Joint Limits
The current implementation uses `sim.model.qpos0` to model sensor/encoder offsets. However, MuJoCo applies joint limits (`jnt_range`) to the raw `qpos` values, not the computed physical position ($q_\text{physical} = q_\text{pos} - q_\text{pos0}$).

As a result, setting a `qpos0` bias effectively shifts the physical workspace of the robot. **This brings a sim-to-real gap!**

# Detailed Description
The current logic assumes:
1.  Encoder Reading (q_pos) = Physical Position (q_phys) + bias (q_pos0)
2.  Physics operates on (q_pos - q_pos0).

However, MuJoCo's limit check is:
range_min <= q_pos <= range_max

Substituting the bias equation:
range_min <= (q_phys + bias) <= range_max

This implies the physical limits are shifted to:
(range_min - bias) <= q_phys <= (range_max - bias)

**Crucially, this means the joint limits act as a hard clamp on the sensor value, not the physical state.** Regardless of the bias, the robot currently always stops when the encoder reads the XML limit value.

# Suggested Fix

When `qpos0` is modified to simulate bias, `sim.model.jnt_range` must be shifted by the same amount to preserve the physical workspace.

```
sim.model.qpos0[joint_id] = bias
sim.model.jnt_range[joint_id, 0] += bias  # Shift min
sim.model.jnt_range[joint_id, 1] += bias  # Shift max
```

**Or should we consider not using qpos0 to simulate the encoder bias?**

kevinzakka:
Fix encoder bias shifting joint limits.

Previously we used MuJoCo's `qpos0` to simulate encoder bias. MuJoCo computes body positions as `(qpos - qpos0)`, so setting `qpos0=bias` makes qpos act like a biased encoder reading while the body stays at the correct physical position.

The problem: qpos has the bias baked in, and joint limits apply to qpos. So the limit triggers based on the biased value, not the true physical position. With bias=0.8 and upper limit=1.0, the limit triggers when qpos hits 1.0, but physical position is only 0.2 at that point. The joint stops 0.8 away from where it should physically be able to go.

This PR adds an encoder_bias field to EntityData instead. Bias is added to observations and subtracted from commands, so limits apply to true physical position.

Code wise, we've made the following changes:
- `EntityData`: added `encoder_bias` field and `joint_pos_biased` property
- `joint_pos_rel`: reads from `joint_pos_biased` if `biased=True`
- `JointPositionAction`: subtracts `encoder_bias` from commands
- Added `randomize_encoder_bias` event function
- Updated tracking task and docs to use `encoder_bias` instead of `qpos0`

Fixes #422.

### DIFF
--- a/src/mjlab/envs/mdp/observations.py
+++ b/src/mjlab/envs/mdp/observations.py
@@ -50,13 +50,15 @@ def projected_gravity(
 
 def joint_pos_rel(
   env: ManagerBasedRlEnv,
+  biased: bool = False,
   asset_cfg: SceneEntityCfg = _DEFAULT_ASSET_CFG,
 ) -> torch.Tensor:
   asset: Entity = env.scene[asset_cfg.name]
   default_joint_pos = asset.data.default_joint_pos
   assert default_joint_pos is not None
   jnt_ids = asset_cfg.joint_ids
-  return asset.data.joint_pos[:, jnt_ids] - default_joint_pos[:, jnt_ids]
+  joint_pos = asset.data.joint_pos_biased if biased else asset.data.joint_pos
+  return joint_pos[:, jnt_ids] - default_joint_pos[:, jnt_ids]
 
 
 def joint_vel_rel(
--- a/docs/api/domain_randomization.md
+++ b/docs/api/domain_randomization.md
@@ -57,7 +57,7 @@ This flag is especially useful when using custom class-based event terms instead
 ## Available Fields
 
 **Joint/DOF:** `dof_armature`, `dof_frictionloss`, `dof_damping`, `jnt_range`,
-`jnt_stiffness`, `qpos0`
+`jnt_stiffness`
 
 **Body:** `body_mass`, `body_ipos`, `body_iquat`, `body_inertia`, `body_pos`,
 `body_quat`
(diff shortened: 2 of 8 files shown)


src/mjlab/envs/mdp/observations.py
```diff
@@ -52,11 +52,12 @@ def joint_pos_rel(
   env: ManagerBasedRlEnv,
   asset_cfg: SceneEntityCfg = _DEFAULT_ASSET_CFG,
 ) -> torch.Tensor:
+  """Joint positions relative to default: joint_pos_biased - default."""
   asset: Entity = env.scene[asset_cfg.name]
   default_joint_pos = asset.data.default_joint_pos
   assert default_joint_pos is not None
   jnt_ids = asset_cfg.joint_ids
-  return asset.data.joint_pos[:, jnt_ids] - default_joint_pos[:, jnt_ids]
+  return asset.data.joint_pos_biased[:, jnt_ids] - default_joint_pos[:, jnt_ids]
```
qiayuanl: For consistency, we might call this function `joint_pos_rel_biased()`, and have another `joint_pos_rel()`.

However, in this case, we need to modify all the task config.

kevinzakka: It's backwards compatible: by default, `encoder_bias` is initialized to 0, so when users aren't using encoder bias randomization, this function returns exactly what it did before.

kevinzakka: joint_pos_rel now takes an arg `biased` that is False by default.

PR #425, opened 2025-12-18
Bump down and pin rsl_rl to 3.1.0.

kevinzakka:
Reason: distributed training is broken.


kevinzakka:
I forgot if we changed anything else at the time of the upgrade 😅

kevinzakka:
Yep we can leave as is, I checked! Thanks @louislelay!

PR #443, opened 2025-12-28
fix: prevent NaNs in `apply_delta_pose` for zero rotation

Bolajiadesina:
- `apply_delta_pose` divided by `angle` when `angle==0`, which can produce NaNs.
- Clamp `angle` before division and add a unit test for zero-rotation delta.


kevinzakka:
Thanks!

PR #462, opened 2026-01-06
Colocalize manager configs and add comprehensive docstrings

kevinzakka:
## Summary

- Move all `*TermCfg` classes from `manager_term_config.py` to their respective manager files (e.g., `ActionTermCfg` → `action_manager.py`)
- Move `ManagerTermBaseCfg` to `manager_base.py` and delete the now-empty `manager_term_config.py`
- Add comprehensive docstrings to all config classes and manager classes
- Unify API: change `commands`/`curriculum` from `| None = None` to `field(default_factory=dict)` for consistency with other manager configs
- Update all `is not None` checks to truthy checks across ~15 files
- Export all config classes from `managers/__init__.py`

## Test plan

- [x] `make check` passes (ruff format, ruff check, ty check, pyright)
- [x] `make test-fast` passes (368 tests)

### DIFF
--- a/scripts/benchmarks/measure_throughput.py
+++ b/scripts/benchmarks/measure_throughput.py
@@ -126,7 +126,7 @@ def benchmark_task(task: str, cfg: ThroughputConfig) -> BenchmarkResult:
   env_cfg.scene.num_envs = cfg.num_envs
 
   # Handle tracking task motion file.
-  if env_cfg.commands:
+  if len(env_cfg.commands) > 0:
     motion_cmd = env_cfg.commands.get("motion")
     if isinstance(motion_cmd, MotionCommandCfg):
       api = wandb.Api()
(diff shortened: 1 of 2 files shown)


scripts/benchmarks/measure_throughput.py
```diff
@@ -126,7 +126,7 @@ def benchmark_task(task: str, cfg: ThroughputConfig) -> BenchmarkResult:
   env_cfg.scene.num_envs = cfg.num_envs
 
   # Handle tracking task motion file.
-  if env_cfg.commands is not None:
+  if env_cfg.commands:
```
brentyi: nit: is it desirable to use truthy checks instead of `len(env_cfg.commands) == 0` ? I'd prefer to be more explicit

kevinzakka: Done, thanks Brent!

PR #468, opened 2026-01-08
Add pre-commit configuration to order the imports

saikishor:
This adds the pre-commit configuration to reorder the imports of the libraries


saikishor:
@kevinzakka you are absolutely right. I missed that part in the pyproject.toml. If this is the case, this PR is not needed anymore. Thanks. I'll go ahead and close this PR

PR #473, opened 2026-01-09
Create variable instead of property to avoid copying tensor everytime.

mktk1117:
The slicing with body_indexes causing the copy of the whole motion tensor every time it is accessing through the properties. This slowed down a lot with large motion_file. This PR fixes it by making a copy in the init.

reference: https://discuss.pytorch.org/t/does-indexing-a-tensor-return-a-copy-of-it/164905


kevinzakka:
Thanks @mktk1117

PR #505, opened 2026-01-19
Show target sphere for all envs in LiftingCommand

siddancha:
## Summary
- [x] Modified `LiftingCommand._debug_vis_impl()` to render target position spheres for all environments instead of only the selected one
- [x] Each sphere now has a unique label (`target_position_{batch}`) to avoid collisions
- [x] Enables visibility of target locations when running multiple parallel environments in both viser and native viewers

### Before

image

### After

image

## Test
- Run `uv run play Mjlab-Lift-Cube-Yam --agent random --num-envs 4 --viewer viser`
- Verify semi-transparent target spheres appear for all 4 environments
- All tests pass

### DIFF
--- a/src/mjlab/tasks/manipulation/mdp/commands.py
+++ b/src/mjlab/tasks/manipulation/mdp/commands.py
@@ -101,17 +101,14 @@ class LiftingCommand(CommandTerm):
     pass
 
   def _debug_vis_impl(self, visualizer: DebugVisualizer) -> None:
-    batch = visualizer.env_idx
-    if batch >= self.num_envs:
-      return
-
-    target_pos = self.target_pos[batch].cpu().numpy()
-    visualizer.add_sphere(
-      center=target_pos,
-      radius=0.03,
-      color=self.cfg.viz.target_color,
-      label="target_position",
-    )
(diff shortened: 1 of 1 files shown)


kevinzakka:
Hi @siddancha,

We've made a deliberate design decision for **debug visualization** to only render for the currently selected env. The reason is that it can become expensive for many environments. I agree with your current change, I just don't know how to balance this change for other debug viz options.

siddancha:
> Hi @siddancha,
> 
> We've made a deliberate design decision for **debug visualization** to only render for the currently selected env. The reason is that it can become expensive for many environments. I agree with your current change, I just don't know how to balance this change for other debug viz options.

@kevinzakka That makes sense. But is there _**some mode**_ where you can visualize sprites for all envs? Something like this still from [mjlab's trailer](https://x.com/i/status/1972757435707424898)? Visualizing all goal sprites seems like an important use case; otherwise it's hard to say if the policy is doing the right thing.

image

kevinzakka:
Hi @siddancha, see #509.

siddancha:
Yayy! Thanks for implementing this feature @kevinzakka!

PR #508, opened 2026-01-19
Fix mocap body visualization in native viewer for batched environments

### LINKED ISSUE #507: Fixed base robots share same base location in multi-env native viewer
## Bug in native viewer

When viewing `Mjlab-Lift-Cube-Yam` in the native viewer with multiple envs, all robots share the same base location instead of separating into a grid layout.

```sh
uv run play Mjlab-Lift-Cube-Yam --wandb-run-path siddancha/mjlab/vjkzne1i --num-envs=4 --viewer=native
```

https://github.com/user-attachments/assets/aa36f255-9c0f-45df-ad55-ce87dbf1861c

## Viser works fine

This problem only occurs in the native viewer, and not in viser:

```sh
uv run play Mjlab-Lift-Cube-Yam --wandb-run-path siddancha/mjlab/vjkzne1i --num-envs=4 --viewer=viser
```

Image

## Potential fix

Not syncing mocap poses from viewer to env [as implemented here](https://github.com/siddancha/mjlab/commit/42063389ef22ec7293c5e38e44f948b9fa0fc7c1) seems to fix this, except that the fixed base is not visualized for all but one env. 

https://github.com/user-attachments/assets/3141dcb9-a318-45ea-8bdc-2b554a581275

kevinzakka:
Remove mocap pose syncing from viewer to environment. With batched environments that use env_origins, the viewer operates in world coordinates while each env has its own spatial offset. Syncing mocap poses back would overwrite correct per-env positions with wrong world-frame values.

Also fix xfrc_applied to update only the selected env_idx rather than broadcasting to all environments.

Requires bumping mujoco to a version that includes the fix for mocap body descendants being incorrectly categorized as static in the visualizer (google-deepmind/mujoco@f3ad0ef).

Fixes #507.

### DIFF
--- a/src/mjlab/viewer/native/viewer.py
+++ b/src/mjlab/viewer/native/viewer.py
@@ -196,7 +196,7 @@ class NativeMujocoViewer(BaseViewer):
       v.sync(state_only=True)
 
   def sync_viewer_to_env(self) -> None:
-    """Copy perturbation forces and mocap poses from viewer to env."""
+    """Copy perturbation forces from viewer to env."""
     if not self.enable_perturbations or self._is_paused or not self.mjd:
       return
     assert self.mjm is not None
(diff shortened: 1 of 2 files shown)


kevinzakka:
cc @siddancha

PR #559, opened 2026-02-02
Adds icon to web pages

louislelay:
I also wanted to use this PR to add the icon to the native viewer on Ubuntu, but I couldn't manage it :sweat:


kevinzakka:
Thanks @louislelay!

PR #561, opened 2026-02-02
Use `set_const` in domain randomization

### LINKED ISSUE #541: Complex domain randomization
Now that https://github.com/google-deepmind/mujoco_warp/commit/0e1040818a0cc21d2b211801e2cc7536cec94818 has been merged, we can implement domain randomization events that require re-computing derived fields (e.g., body_mass).

oxkitsune:
### What

When randomizing certain model fields (e.g., `body_mass`, `body_inertia`, etc.), the sim has derived values like `body_subtreemass` and `body_invweight0` that currently aren't updated. 

google-deepmind/mujoco_warp@0e10408 exposed `set_const`, which recomputes the derived values for these fields.

I added the `requires_recompute` property to `FieldSpec`. For fields with this flag we can then automatically call `set_constant`. I marked the fields as `requires_recompute` based on the table in mjwarp's `set_constant` docstring:
https://github.com/google-deepmind/mujoco_warp/blob/75280040d31db4a68a7d4b9bbb1cf5f5820ccc93/mujoco_warp/_src/io.py#L1880-L1894

These derived model fields (like `body_subtreemass`, `body_invweight0`) must also be expanded for each environment, just like the domain randomization fields. Without expansion, `set_const` would write the recomputed values to a shared array with incorrect memory layout, since it expects per-env storage but would be writing to a single shared buffer.                      

Note: Currently I'm just called `set_const`. However, not all fields require everything to be recomputed, we could also call one of the `set_const_0` or `set_const_fixed`  sub-functions directly. This would make the code a bit more complex, and I think this is neat 😄 

TODO:

- [x] measure how much this influences performance
- [x] make sure this doesn't break any existing environment

Should hopefully close #541

### DIFF
--- a/src/mjlab/envs/manager_based_rl_env.py
+++ b/src/mjlab/envs/manager_based_rl_env.py
@@ -262,7 +262,28 @@ class ManagerBasedRlEnv:
     self.event_manager = EventManager(self.cfg.events, self)
     print_info(f"[INFO] {self.event_manager}")
 
-    self.sim.expand_model_fields(self.event_manager.domain_randomization_fields)
+    # Expand per-env model fields that may be modified by domain randomization.
+    # Also expand derived model fields, since set_const recomputes them with
+    # per-env values; without expansion, it writes to a shared array with wrong
+    # memory layout.
+    derived_model_fields = {
+      "body_subtreemass",
+      "tendon_length0",
+      "dof_invweight0",
+      "body_invweight0",
+      "tendon_invweight0",
+      "cam_pos0",
+      "cam_poscom0",
+      "cam_mat0",
+      "light_pos0",
+      "light_poscom0",
+      "light_dir0",
+      "actuator_acc0",
+    }
+    fields_to_expand = (
+      set(self.event_manager.domain_randomization_fields) | derived_model_fields
+    )
+    self.sim.expand_model_fields(tuple(fields_to_expand))
 
     # Command manager (must be before observation manager since observations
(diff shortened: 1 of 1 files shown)


oxkitsune:
Ah, it's a bit more complicated than I anticipated. Some extra model fields need to be expanded, pushing something in a sec.

src/mjlab/envs/manager_based_rl_env.py
```diff
@@ -262,7 +262,28 @@ def load_managers(self) -> None:
     self.event_manager = EventManager(self.cfg.events, self)
     print_info(f"[INFO] {self.event_manager}")
 
-    self.sim.expand_model_fields(self.event_manager.domain_randomization_fields)
+    # Expand per-env model fields that may be modified by domain randomization.
+    # Also expand derived model fields, since set_const recomputes them with
+    # per-env values; without expansion, it writes to a shared array with wrong
+    # memory layout.
+    derived_model_fields = {
```
oxkitsune, line 269: Perhaps this would be better off as a constant somewhere.

kevinzakka:
Hi @oxkitsune, in my (growing) backlog of mjlab improvements, I want to revamp the domain randomization system because it's a little ugly and hacky at the moment. Gonna hold off on reviewing and merging this PR until I have come up with a decent plan. Can share it with you when it's ready.

oxkitsune:
Thanks for the heads up!

PR #583, opened 2026-02-05
Fixes ignored evaluation ranges in G1 flat tasks

louislelay:
The velocity ranges set for evaluation of G1 flat tracking task are no longer overwritten by the curriculum upon reset. (and added the same to Go1)


louislelay:
Thanks! :)

PR #589, opened 2026-02-06
Upgrade rsl-rl-lib to 4.0.0 and fixes breaking changes

louislelay:
This PR upgrades the `rsl-rl-lib` dependency from **3.1.0** to **4.0.0** and resolves all associated breaking changes across the repository. The update migrates the configuration architecture from a single `RslRlPpoActorCriticCfg` to independent `actor` and `critic` definitions using the new `RslRlModelCfg` class. To align with the updated library API and eliminate warnings, all observation groups previously named `policy` are now renamed to `actor`.

To prevent breaking existing workflows, I implemented a migration layer in the runner that automatically maps legacy `model_state_dict` keys to the new format when loading older checkpoints. Additionally, I added missing type annotations to the runner, and synchronized the entire test suite, task configs, and notebooks with the new library standards.

### DIFF
--- a/tests/test_runner.py
+++ b/tests/test_runner.py
@@ -6,6 +6,7 @@ from pathlib import Path
 
 import mujoco
 import pytest
+import torch
 from conftest import get_test_device
 
 from mjlab.actuator import XmlMotorActuatorCfg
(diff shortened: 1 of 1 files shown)


louislelay:
Thanks for the review! 🥳🚀

PR #596, opened 2026-02-07
Add MetricsManager for custom metric logging

### LINKED ISSUE #584: Approach to inject and inspect new metrics during training
Hello!

First of all, thanks for all the great work that you guys put in here. I've been using mjlab for a couple of weeks, it is very modular and easy to integrate. 

During my development, I stumbled upon the need to get the metrics of different data now and then. Right now, the metrics are recorded at different places of the rewards or command, etc. Whenever I require a specific metric, I have to inject the metrics logging into a new reward or add it to the already existing rewards, and then rerun my training to introspect it via wandb or tensorboard.

Upon going through this for quite some iterations, I'm wondering if there is a need for a Metrics Manager that can log certain metrics of users' interest without having to touch the existing reward functions etc. What do you think about it? (or) Do you suggest simply adding a reward function, but with a weight equal to zero?

I would love to hear your opinion before starting to develop things 

Thank you

kevinzakka:
## Summary

Adds a `MetricsManager` so users can log custom per-step metrics during training without hacking reward functions or adding zero-weight reward terms. Closes #584.

- **New**: `MetricsManager`, `MetricsTermCfg`, `NullMetricsManager` in `managers/metrics_manager.py`
- **Integration**: wired into `ManagerBasedRlEnv` config, `step()`, and `_reset_idx()`
- Terms use the same callable signature as rewards `(env, **params) → Tensor[num_envs]`
- No weight, no dt scaling — metrics are observational, not reward signals
- Episode values are true per-step averages (`sum / step_count`), so a metric in [0,1] stays in [0,1] in wandb
- Empty config means zero overhead (`NullMetricsManager`)

### Example usage

```python
  from mjlab.managers import MetricsTermCfg

  def joint_velocity_magnitude(env, asset_cfg):
    """L1 norm of joint velocities."""
    return env.scene[asset_cfg.name].data.joint_vel.abs().sum(dim=-1)  # (num_envs,)

  @dataclass(kw_only=True)
  class MyEnvCfg(ManagerBasedRlEnvCfg):
    metrics = {
      "joint_vel_mag": MetricsTermCfg(
        func=joint_velocity_magnitude,
        params={"asset_cfg": SceneEntityCfg(name="robot")},
      ),
    }
```

On episode reset, `Episode_Metrics/joint_vel_mag` appears in `extras["log"]` and flows to wandb/tensorboard automatically.

## Test plan

- [x] `uv run pytest tests/test_metrics_manager.py` — 6 targeted tests
- [x] `uv run pytest tests/test_rewards.py` — no regression
- [x] `uv run ty check` / `uv run pyright` — clean
- [x] `uv run ruff check && uv run ruff format` — clean


brentyi (commented):
+100, seems super useful!

I'm wondering if any if these things are possible in a simple/not overengineered way:

Does it make sense to allow customization of how metrics are "reduced" between steps and before logging? In the extreme case it'd be nice, for example, to be able to specify logging for std of metrics, stds of per-episode means, histograms, histograms of per-episode stds, etc.

Can we implement any of the existing things that are logged (eg rewards) as default terms in the metrics manager?

src/mjlab/envs/manager_based_rl_env.py
```diff
@@ -291,6 +299,11 @@ def load_managers(self) -> None:
     else:
       self.curriculum_manager = NullCurriculumManager()
     print_info(f"[INFO] {self.curriculum_manager}")
+    if len(self.cfg.metrics) > 0:
+      self.metrics_manager = MetricsManager(self.cfg.metrics, self)
+    else:
+      self.metrics_manager = NullMetricsManager()
```
brentyi, line 309: Perhaps this can just be `self.metrics_manager = None` and we can add if guards?

src/mjlab/managers/metrics_manager.py
```diff
@@ -0,0 +1,144 @@
...
+    for term_cfg in self._class_term_cfgs:
+      term_cfg.func.reset(env_ids=env_ids)
+    return extras
+
+  def compute(self) -> None:
+    self._step_count += 1
+    for term_idx, (name, term_cfg) in enumerate(
+      zip(self._term_names, self._term_cfgs, strict=False)
+    ):
+      value = term_cfg.func(self._env, **term_cfg.params)
```
brentyi, line 97: are all metrics always computed every step?

src/mjlab/envs/manager_based_rl_env.py
kevinzakka: Went with the `Null` object to stay consistent with `NullCurriculumManager` / `NullCommandManager`. It's the same pattern, and it keeps `step()` and `_reset_idx()` free of `if is not None` guards.

src/mjlab/managers/metrics_manager.py
kevinzakka: This mimics the reward structure. If a user has an expensive metric they only want occasionally, they can gate it inside their function. I'm open to adding a `compute_interval` later if there's demand?

kevinzakka:
Thanks for the review @brentyi!

> Does it make sense to allow customization of how metrics are "reduced" between steps and before logging?

The rsl_rl logger only supports scalars. Everything in `extras["log"]` goes through `torch.mean()` then `add_scalar()` ([logger.py:171](https://github.com/mujocolab/rsl_rl/blob/main/rsl_rl/utils/logger.py#L171)). So histograms/distributions aren't possible without changing the logger. We could try submitting a PR to rsl_rl. In the meantime, for std or other reductions, users can just add a second metric term that computes it directly (e.g. `joint_vel_std` alongside `joint_vel_mean`). Both log as scalars and it works today with no extra machinery.

> Can we implement any of the existing things that are logged (eg rewards) as default terms in the metrics manager?

In principle yes. Note however that rewards use dt-scaled sums divided by `max_episode_length_s` while metrics use raw sums divided by `step_count`. We'd have to add machinery to the metric manager to support this.

src/mjlab/envs/manager_based_rl_env.py
brentyi: Ah I see I remember discussing this last year, still mixed feelings about it but consistency is good 👍

src/mjlab/managers/metrics_manager.py
brentyi: Makes sense! Seems easy to add later if needed [in a compatible way]

brentyi:
> In the meantime, for std or other reductions, users can just add a second metric term that computes it directly (e.g. `joint_vel_std` alongside `joint_vel_mean`). Both log as scalars and it works today with no extra machinery.

Makes sense!

To check my understanding: we wouldn't be able to reuse intermediates right? And we could compute std across episodes within a single timestep but not across timesteps within an episode? These are a bit annoying but fine.

> Note however that rewards use dt-scaled sums divided by max_episode_length_s while metrics use raw sums divided by step_count.

Makes sense. It seems kind of nice to consolidate logic but I don't feel strongly about this.

src/mjlab/managers/metrics_manager.py
saikishor: I agree, if there is a need, we can add later

brentyi (approved):
👍

PR #597, opened 2026-02-07
Add camera sensor, unified sensor context, and spatial-softmax CNN model

kevinzakka:
Introduce CameraSensor alongside the existing RaycastSensor, backed by a shared SensorContext that manages the mujoco_warp RenderContext.  Add a SpatialSoftmaxCNNModel (drop-in replacement for rsl_rl's CNNModel) that replaces global pooling with spatial soft-argmax keypoints.  Wire up the Yam D405 camera and vision-based lift-cube tasks end-to-end.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ dependencies = [
   "viser>=1.0.21",
   "moviepy",
   "tensordict",
-  "rsl-rl-lib==4.0.0",
+  "rsl-rl-lib==4.0.1",
   "tensorboard>=2.20.0",
   "onnxscript>=0.5.4",
   "wandb>=0.22.3",
@@ -121,8 +121,7 @@ explicit = true
 warp-lang = { index = "nvidia", marker = "sys_platform != 'darwin'" }
 mujoco = { index = "mujoco" }
 torch = { index = "pytorch-cu128", extra = "cu128", marker = "sys_platform != 'darwin'" }
-mujoco-warp = { git = "https://github.com/google-deepmind/mujoco_warp", rev = "6f5ba7dd99d73516543aef99e52916b9c71b1567" }
-rsl-rl-lib = { git = "https://github.com/kevinzakka/rsl_rl", rev = "fix/cnn-model-init-order" }
+mujoco-warp = { git = "https://github.com/google-deepmind/mujoco_warp", rev = "b75f85b4dbc60f2d6e9277d345953aa1f1dd1611" }
 
 [tool.ruff]
 src = ["src"]  # Helpful for recognizing first-party imports.
--- a/src/mjlab/rl/spatial_softmax.py
+++ b/src/mjlab/rl/spatial_softmax.py
@@ -176,13 +176,13 @@ class SpatialSoftmaxCNNModel(CNNModel):
           **group_cfg,
         )
 
-    self._cnn_latent_dim = 0
+    self.cnn_latent_dim = 0
     for cnn in _cnns.values():
       if cnn.output_channels is not None:
         raise ValueError(
           "The output of the CNN must be flattened before passing it to the MLP."
         )
-      self._cnn_latent_dim += int(cnn.output_dim)  # type: ignore[arg-type]
+      self.cnn_latent_dim += int(cnn.output_dim)  # type: ignore[arg-type]
 
     MLPModel.__init__(
       self,
--- a/uv.lock
+++ b/uv.lock
@@ -1298,10 +1298,10 @@ docs = [
 requires-dist = [
     { name = "moviepy" },
     { name = "mujoco", specifier = ">=3.4.0", index = "https://py.mujoco.org/" },
-    { name = "mujoco-warp", git = "https://github.com/google-deepmind/mujoco_warp?rev=6f5ba7dd99d73516543aef99e52916b9c71b1567" },
+    { name = "mujoco-warp", git = "https://github.com/google-deepmind/mujoco_warp?rev=b75f85b4dbc60f2d6e9277d345953aa1f1dd1611" },
     { name = "onnxscript", specifier = ">=0.5.4" },
     { name = "prettytable" },
-    { name = "rsl-rl-lib", git = "https://github.com/kevinzakka/rsl_rl?rev=fix%2Fcnn-model-init-order" },
+    { name = "rsl-rl-lib", specifier = "==4.0.1" },
     { name = "tensorboard", specifier = ">=2.20.0" },
     { name = "tensordict" },
     { name = "torch", specifier = ">=2.7.0" },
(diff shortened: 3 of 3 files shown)


src/mjlab/envs/mdp/observations.py
```diff
@@ -104,3 +104,28 @@ def builtin_sensor(env: ManagerBasedRlEnv, sensor_name: str) -> torch.Tensor:
...
+
+  Returns:
+    Tensor of shape [B, N] where B is num_envs and N is num_rays.
+  """
+
+  sensor: RayCastSensor = env.scene[sensor_name]
+  heights = (
+    sensor.data.pos_w[:, 2].unsqueeze(1) - sensor.data.hit_pos_w[..., 2] - offset
+  )
+  heights[sensor.data.distances < 0] = float("-inf")
```
StafaH: The convention across raycast/render is that no-hits are -1, might be better if we keep that convention. Downstream masking should just use `>= 0` to find which rays have hit

src/mjlab/sensor/sensor_context.py
```diff
@@ -0,0 +1,303 @@
...
+      for s in self.camera_sensors:
+        cam_active[s.camera_idx] = True
+        cam_res.append((s.cfg.width, s.cfg.height))
+        render_rgb.append("rgb" in s.cfg.data_types)
+        render_depth.append("depth" in s.cfg.data_types)
+    else:
+      # Raycasts-only: need BVH but no camera rendering.
+      use_textures = False
+      use_shadows = False
+      enabled_geom_groups = [0, 1, 2, 3, 4, 5]
```
StafaH: This will force raycast sensors to cast against geoms (visual and collision). This needs to come from config and not be all by default (0/3 or 0/1/2 are better defaults)

Also along the same lines, we should check that raycast_enabled_geoms == render_enabled_geoms for when there are both types of sensors and maybe raise a warning before using a (e.g. `set(raycast_geoms + render_geoms)`)

src/mjlab/sensor/raycast_sensor.py
```diff
@@ -711,29 +703,54 @@ def _perform_raycast(self) -> None:
...
     self._normals_w = wp.to_torch(self._ray_normal).view(num_envs, self._num_rays, 3)
     self._distances[self._distances > self.cfg.max_distance] = -1.0
 
-    # Compute hit positions: origin + direction * distance.
-    # For misses (distance = -1), hit_pos_w will be invalid (but normals_w are zero).
-    assert self._distances is not None
     hit_mask = self._distances >= 0
-    hit_pos_w = world_origins.clone()
-    hit_pos_w[hit_mask] = world_origins[hit_mask] + world_rays[
+    hit_pos_w = torch.zeros_like(self._cached_world_origins)
```
StafaH: Is it better to keep the original `hit_pos_w = self._cached_world_origins.clone()`? as zeros would mean the ray hit the world origin, vs ray_origin which might be more interpretable.

src/mjlab/scene/scene.py
```diff
@@ -133,6 +144,21 @@ def initialize(
     for sensor in self._sensors.values():
       sensor.initialize(mj_model, model, data, self._device)
 
+    # Create SensorContext if any camera or raycast sensors exist.
+    camera_sensors = [s for s in self._sensors.values() if isinstance(s, CameraSensor)]
```
StafaH: This could be cumbersome to maintain, can we just add a field to sensors that says `requires_sensor_context: bool = False`? Also it might be possible for other sensors (distance?) to eventually use rctx internally.

src/mjlab/sensor/camera_sensor.py
```diff
@@ -0,0 +1,201 @@
...
+    return self._camera_idx
+
+  def edit_spec(
+    self,
+    scene_spec: mujoco.MjSpec,
+    entities: dict[str, Entity],
+  ) -> None:
+    del entities
+
+    if self._is_wrapping_existing:
```
StafaH, line 153: What about the other fields (resolution, output, projection) that were added to the xml/spec recently? We should properly resolve all fields.

src/mjlab/sensor/sensor_context.py
```diff
@@ -0,0 +1,303 @@
...
+
+    if cam_idx not in self._cam_idx_to_list_idx:
+      available = list(self._cam_idx_to_list_idx.keys())
+      raise KeyError(
+        f"Camera ID {cam_idx} not found in SensorContext. "
+        f"Available camera IDs: {available}"
+      )
+
+    list_idx = self._cam_idx_to_list_idx[cam_idx]
+    if not self._render_rgb[list_idx]:
```
StafaH: The convention internally in the renderer is that `(rgb/depth)_adr = -1` for not enabled cameras, you can use that as a check although its not necessary.

src/mjlab/asset_zoo/robots/i2rt_yam/xmls/yam.xml
```diff
@@ -127,6 +130,14 @@
                       </body>
                     </body>
                   </body>
+                  <body name="camera_d405" pos="-0.0107 0.079729 0.066021" quat="0.953717 0.3007058 0 0">
+                    <geom class="visual" mesh="camera_d405"/>
```
StafaH, line 134: If the rays spawn from inside this geom, you won't see this geom since the inside faces of a mesh aren't rendered, but your throughput would be substantially lower (since each ray needs to do a ray-mesh of something that cannot be rendered, and ray-mesh is expensive). Just something to keep in mind.

src/mjlab/envs/mdp/observations.py
kevinzakka: Done.

src/mjlab/sensor/sensor_context.py
kevinzakka: I made the default (0, 1, 2) and None means use all.

src/mjlab/scene/scene.py
kevinzakka: Done.

src/mjlab/sensor/raycast_sensor.py
kevinzakka: Changed to ray_origin.

src/mjlab/sensor/camera_sensor.py
kevinzakka: Done.

src/mjlab/sensor/sensor_context.py
kevinzakka: Done.

src/mjlab/asset_zoo/robots/i2rt_yam/xmls/yam.xml
kevinzakka: It's in group 4 and we only render groups (0, 1, 2) by default and specifically in the case of the YAM lift cube config, we only render groups (0, 3).

kevinzakka:
Need to wait until https://github.com/leggedrobotics/rsl_rl/pull/175 lands.

kevinzakka:
All done! @StafaH shall I merge?

louislelay:
🥳🎉🎉

PR #599, opened 2026-02-07
Consolidate two sim.forward() calls into one before observations

kevinzakka:
After the decimation loop, mj_step leaves derived quantities (xpos, xquat, site_xpos, cvel, sensordata) stale by one physics substep. The previous fix (4d89539) added an unconditional forward() after the decimation loop plus a conditional one after the reset block, resulting in two calls per step, sometimes back to back.

Replace both with a single unconditional forward() placed after the reset block and before observation computation. This serves all envs in one pass: non-reset envs get post-decimation kinematics, reset envs get post-reset kinematics.

The tradeoff is that termination and reward managers now see derived quantities that lag qpos/qvel by one physics substep. This staleness is consistent across all envs and all steps, so the MDP remains well-defined. Observations, which the policy acts on, are always fresh.


yuvaltassa:
This is only for resets right? It's not like every step is now step+forward... correct?

PR #605, opened 2026-02-09
Propage and reuse the num_steps_per_env information

saikishor:
When launching several trainings via command line, when trying to set the curriculum steps, and also setting the num_steps_per_env, it felt a bit messy to calculate the corresponding value and use it. Instead, I propose to propagate this parameter to the env and then reuse it within the curriculum, by defining the iteration instead of step, this would make it easier to modify it at runtime

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -64,8 +64,8 @@ This is our recommended way to use ``mjlab``. You have
 your own project and want to use ``mjlab`` as a dependency
 using ``uv``.
 
-Install uv
-^^^^^^^^^^
+1. Install uv
+^^^^^^^^^^^^^
 
 If you do not have ``uv`` installed, run:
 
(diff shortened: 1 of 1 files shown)


saikishor:
> Hey @saikishor, run `make test-all` please! 

Yes, I'm on it already

kevinzakka:
Hi @saikishor, thanks for tackling this. Agreed that specifying things in terms of iterations (N steps) is nicer than steps (one `env.step()` call).

I'm not thrilled with the current approach though: it propagates `num_steps_per_env` to both `train.py` and `runner.py`, and copy-pastes the same iteration->step conversion + validation into three curriculum functions. That's a lot of plumbing for what's essentially a config convenience.

I put up #606 as an alternative. It resolves "iteration"->"step" once at config time (before the env is created), so curriculum functions don't need to change at all. Take a look and let me know what you think.

saikishor:
Closing this in favour of #606

PR #606, opened 2026-02-10
Allow curriculum stages to be defined in iterations instead of raw steps

kevinzakka:
Alternative to #605. Curriculum configs can now use `"iteration"` instead of `"step"` in stage dicts (e.g. `{"iteration": 5000}` vs `{"step": 5000 * 24}`), which is more intuitive and doesn't require knowing `num_steps_per_env`.

Rather than propagating `num_steps_per_env` into the env config and duplicating conversion logic across curriculum functions, this resolves `"iteration"` → `"step"` once at config time in `train.py` before the env is created. Curriculum functions remain unchanged.

### DIFF
--- a/tests/test_curriculum_manager.py
+++ b/tests/test_curriculum_manager.py
@@ -1,0 +1,99 @@
+"""Tests for resolve_curriculum_iterations()."""
+
+import copy
+
+import pytest
+
+from mjlab.managers.curriculum_manager import (
+  CurriculumTermCfg,
+  resolve_curriculum_iterations,
+)
+
+
+def _dummy_func(*args, **kwargs):
+  pass
+
+
+def _make_curriculum(stages_key, stages):
+  """Build a minimal curriculum dict with one term."""
+  return {
+    "term": CurriculumTermCfg(
+      func=_dummy_func,
+      params={"some_name": "twist", stages_key: stages},
+    )
+  }
+
+
+def test_converts_iteration_to_step():
+  stages = [
+    {"iteration": 0, "lin_vel_x": (-1.0, 1.0)},
+    {"iteration": 5000, "lin_vel_x": (-2.0, 2.0)},
+  ]
+  curriculum = _make_curriculum("velocity_stages", stages)
+  resolve_curriculum_iterations(curriculum, num_steps_per_env=24)
+
+  resolved = curriculum["term"].params["velocity_stages"]
+  assert resolved[0] == {"step": 0, "lin_vel_x": (-1.0, 1.0)}
+  assert resolved[1] == {"step": 120000, "lin_vel_x": (-2.0, 2.0)}
+  assert all("iteration" not in stage for stage in resolved)
+
+
+def test_leaves_step_based_stages_untouched():
+  stages = [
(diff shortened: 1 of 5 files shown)


tests/test_curriculum_manager.py
```diff
@@ -0,0 +1,99 @@
...
+    {"iteration": 5000, "lin_vel_x": (-2.0, 2.0)},
+  ]
+  curriculum = _make_curriculum("velocity_stages", stages)
+  resolve_curriculum_iterations(curriculum, num_steps_per_env=24)
+
+  resolved = curriculum["term"].params["velocity_stages"]
+  assert resolved[0] == {"step": 0, "lin_vel_x": (-1.0, 1.0)}
+  assert resolved[1] == {"step": 120000, "lin_vel_x": (-2.0, 2.0)}
+  assert "iteration" not in resolved[0]
+  assert "iteration" not in resolved[1]
```
saikishor:
```suggestion
  assert all("iteration" not in stage for stage in resolved)
```

tests/test_curriculum_manager.py
```diff
@@ -0,0 +1,99 @@
...
+  from mjlab.tasks.velocity.velocity_env_cfg import make_velocity_env_cfg
+
+  env_cfg = make_velocity_env_cfg()
+  resolve_curriculum_iterations(env_cfg.curriculum, num_steps_per_env=24)
+
+  stages = env_cfg.curriculum["command_vel"].params["velocity_stages"]
+  assert stages[0]["step"] == 0
+  assert stages[1]["step"] == 5000 * 24
+  assert stages[2]["step"] == 10000 * 24
+  assert "iteration" not in stages[0]
```
saikishor:
```suggestion
  assert all("iteration" not in stage for stage in stages)
```

tests/test_curriculum_manager.py
```diff
@@ -0,0 +1,99 @@
...
+    make_lift_cube_env_cfg,
+  )
+
+  env_cfg = make_lift_cube_env_cfg()
+  resolve_curriculum_iterations(env_cfg.curriculum, num_steps_per_env=24)
+
+  stages = env_cfg.curriculum["joint_vel_hinge_weight"].params["weight_stages"]
+  assert stages[0]["step"] == 0
+  assert stages[1]["step"] == 500 * 24
+  assert stages[2]["step"] == 1000 * 24
```
saikishor, line 98:
```suggestion
  assert stages[2]["step"] == 1000 * 24
  assert all("iteration" not in stage for stage in stages)
```

tests/test_curriculum_manager.py
kevinzakka: Done.

tests/test_curriculum_manager.py
kevinzakka: Done.

kevinzakka:
This broke a lot of things so reverting until I have a clean non-breaking change.

saikishor:
> This broke a lot of things so reverting until I have a clean non-breaking change.

Got it.  Thanks for updating here. So, are parts of play and train broken?. If you can let me know. I'll be happy to take a look

PR #622, opened 2026-02-12
Adding domain randomization support for special Tendon

samindaa:
To add domain randomization for Tendon based models.

Supports: ``tendon_armature``, ``tendon_damping``, and ``tendon_frictionloss``

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - 'v*'
 
 permissions:
   contents: write
@@ -16,6 +18,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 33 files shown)


louislelay (approved):
LGTM! @kevinzakka wdyt? I know you're planning a DR revamp butit might be worth merging this in the meantime.

samindaa (commented):
Removed tendon_length0 per recommendation

samindaa:
> Thanks for adding tendon randomization support! However, there's a critical issue: `tendon_length0` should not be randomized directly because it's a computed/derived field in MuJoCo, not a direct input parameter.
> 
> According to the `set_const` [documentation](https://github.com/google-deepmind/mujoco_warp/blob/7c71b02cd02494b096c36d6b6327027ec960b7d6/mujoco_warp/_src/io.py#L1960) in mujoco_warp, `tendon_length0` is computed from the tendon path geometry at `qpos0` and is not listed as a safe field to modify. The correct approach would be to randomize `qpos0` and call `set_const` to recompute derived fields.
> 
> For the other fields (`tendon_armature`, `tendon_damping`, `tendon_frictionloss`), these are direct physical parameters, but there's a subtlety: `tendon_damping` has a caveat that if you change it across the zero boundary (from zero to non-zero or vice versa), you need to call `set_const`. Your current test is safe because it scales 1.0 by [0.5, 1.5] which stays non-zero, but users randomizing tendons that start at zero could hit issues.
> 
> Given these complexities, I think we need a more robust domain randomization system that handles `set_const` calls transparently. Would you be open to removing `tendon_length0` from this PR and I can revisit tendon randomization after improving the DR infrastructure?

+1

samindaa:
Thank you Louis and Kevin for detailed explanations. All the recommendations have been applied.

louislelay (commented):
Nice! I reopened your PR but I'll let Kevin do the review.

samindaa (commented):
doc correction

PR #625, opened 2026-02-13
Extend Viser plotting to plot metrics

saikishor:
With the following change:

If the metrics are defined, it will add a tab, and the user will be able to plot them

image

If the metrics are not defined, it should bethe  same as the old behaviour:

image


src/mjlab/viewer/viser/term_plotter.py
```diff
@@ -49,7 +51,7 @@ def _(_) -> None:
       for handle in self._plot_handles.values():
         handle.visible = self._enabled_checkbox.value
 
-    # Create individual plot for each reward term
+    # Create individual plot for each term
     for name in self._term_names:
```
kevinzakka: Nit: name here shadows the name parameter. Consider `for term_name in self._term_names:` to avoid confusion.

saikishor: Done

kevinzakka (approved):

PR #626, opened 2026-02-13
fix(play): use MjlabOnPolicyRunner as default runner

griffinaddison:
## Summary
- `play.py` was falling back to the base rsl-rl `OnPolicyRunner` when no custom `runner_cls` was registered for a task
- `MjlabOnPolicyRunner` strips `cnn_cfg: None` from the config before passing it to rsl-rl, preventing `MLPModel` from receiving an unexpected keyword argument
- This fixes a `TypeError: MLPModel.__init__() got an unexpected keyword argument 'cnn_cfg'` when playing any task that uses the `RslRlModelCfg` API

## Test plan
- [x] `uv run play Mjlab-Lift-Cube-Yam --wandb-run-path <run>` no longer crashes with `cnn_cfg` error

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -14,6 +14,14 @@ Added
   soft joint-limit avoidance, and null-space posture regularization.
   Includes an interactive viser demo (``scripts/demos/differential_ik.py``).
 
+Fixed
+^^^^^
+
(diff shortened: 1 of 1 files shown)


griffinaddison:
Thank you!

PR #630, opened 2026-02-13
Fix camera tracking when toggling certain options

saikishor:
I've observed that when trying to visualize contacts or contact forces, the camera tracking is broken once I try to visualize them on the native viewer, this PR aims to fix that by enforcing to continue tracking in case it is broken by certain options

### DIFF
--- a/src/mjlab/viewer/viser/__init__.py
+++ b/src/mjlab/viewer/viser/__init__.py
@@ -26,6 +26,6 @@ from mjlab.viewer.viser.conversions import (
 from mjlab.viewer.viser.conversions import (
   rotation_quat_from_vectors as rotation_quat_from_vectors,
 )
-from mjlab.viewer.viser.reward_plotter import ViserRewardPlotter as ViserRewardPlotter
 from mjlab.viewer.viser.scene import ViserMujocoScene as ViserMujocoScene
+from mjlab.viewer.viser.term_plotter import ViserTermPlotter as ViserTermPlotter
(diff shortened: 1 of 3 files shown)


saikishor:
**Current state:**


https://github.com/user-attachments/assets/acc4baf5-baba-41f3-9476-99d169e6c90d

**With the proposed fix:**


https://github.com/user-attachments/assets/4f121c81-9d06-4ddf-9f72-9d69541a4537

kevinzakka:
Hi @saikishor, it's a longstanding mujoco bug. I've made [a PR](https://github.com/google-deepmind/mujoco/pull/3108) to address it.

saikishor:
> Hi @saikishor, it's a longstanding mujoco bug. I've made [a PR](https://github.com/google-deepmind/mujoco/pull/3108) to address it.

Got it. Thank you for fixing it upstream. Just curious, should we merge this temporarily as a patch or do you prefer to close it?

kevinzakka:
If you're using mjlab at HEAD, you'll get this feature the day it lands since we use nightly mujoco builds. So I'd prefer we just wait instead of adding then deleting code.

saikishor:
> If you're using mjlab at HEAD, you'll get this feature the day it lands since we use nightly mujoco builds. So I'd prefer we just wait instead of adding then deleting code.

Sure thing 🫡 
Thanks again

saikishor:
For future reference:

Fixed by https://github.com/mujocolab/mjlab/pull/774

PR #632, opened 2026-02-14
Add differential IK action space.

### LINKED ISSUE #452: Cartesian Action Space
Hi,

Thanks for the nice work! I tried the manipulation tasks and found that it uses `JointPositionActionCfg`. I'm wondering are there any existing code or future plan to support cartesian action (6D pose) for RL training. Or, are there any existing tools that supports batched IK so that we can use along with `mjlab`? Any suggestiosn on this would be greatly appreciated!

Thanks!

### LINKED ISSUE #598: Unable to get site jacobians for implementing diff-ik/osc type controllers.
I wanted to implement diff-ik/osc type controllers. I am not able to find something like site_jacobian etc, which can give jacobian of a robot.

Any idea how can I do it.

kevinzakka:
Add a differential IK action term that converts task-space commands into joint position targets via damped least-squares, applied each decimation substep.

Three command modes are supported: position-only (3D), relative delta pose as position + axis-angle (6D), and absolute pose as position + quaternion (7D). Additionally, joint limit avoidance and posture regularization can optionally bias the solver away from limits or toward a target configuration.

The end-effector can be a body, site, or geom.

Fixes #598.
Fixes #452.

https://github.com/user-attachments/assets/c5c8b382-8931-4bde-8bc3-dbee5e3068d2

### DIFF
--- a/src/mjlab/envs/mdp/actions/differential_ik.py
+++ b/src/mjlab/envs/mdp/actions/differential_ik.py
@@ -50,6 +50,13 @@ class DifferentialIKActionCfg(ActionTermCfg):
   """If True, actions are deltas applied to the current frame pose.
   If False, actions are absolute targets."""
 
+  scale: float = 1.0
+  """Scaling factor applied to actions in relative mode.
+
+  Ignored in absolute mode. In relative mode, maps raw policy outputs
+  to task-space deltas (meters for position, radians for orientation).
+  """
+
   damping: float = 0.05
   """Damping coefficient (lambda) for the DLS pseudoinverse."""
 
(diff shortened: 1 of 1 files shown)


src/mjlab/envs/mdp/actions/differential_ik.py
```diff
@@ -0,0 +1,309 @@
...
+  """Element type of the end-effector."""
+
+  ee_name: str
+  """Name of the end-effector element on the entity."""
+
+  use_relative_mode: bool = True
+  """If True, actions are deltas applied to the current EE pose.
+  If False, actions are absolute targets."""
+
+  scale: float = 1.0
```
brentyi: should there be separate scales for position / orientation?

kevinzakka: Removed it. For joint position actions, scale has a physical meaning (it maps policy outputs to achievable joint range via effort/stiffness), but for task space commands there's no equivalent. In absolute mode you'd be scaling a pose which is nonsensical, and in relative mode the policy can just learn the right output magnitude. The solver weights already handle the relative emphasis between position and orientation.

kevinzakka:
Trained a lift cube policy and it did as well as joint action policy but required I add back the scale parameter. Satisfied with the result so going to merge.

PR #637, opened 2026-02-15
Replace moviepy with mediapy for video recording

kevinzakka:
## Summary
Replaces `moviepy` with `mediapy` for video recording. `mediapy` handles cloud storage paths (GCS, S3) natively by writing to a temp file and copying to the remote destination, so `VideoRecorder` now supports remote `video_folder` paths without spreading `etils.epath` across the codebase. `video_folder` now accepts `str | Path`. Supersedes #555.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -29,6 +29,8 @@ Changed
   from ``FIELD_SPECS`` in domain randomization. These fields have derived
   quantities that require ``set_const`` to recompute; without that call,
   randomizing them silently breaks physics.
+- Replaced ``moviepy`` with ``mediapy`` for video recording. ``mediapy``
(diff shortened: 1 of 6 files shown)


rdeits-bd:
I'm not sure if this was intended, but the previous `moviepy` implementation used https://github.com/imageio/imageio-ffmpeg to provide its own ffmpeg. With this change, we're relying on system `ffmpeg` existing, which requires some extra work for weird non-docker use-cases like ours. It looks like the tests are also skipped now when system `ffmpeg` is missing, which also seems inconvenient. 

What do you think about restoring the `imageio-ffmpeg` dependency just to provide a stable copy of ffmpeg rather than whatever random copy is on the system?

PR #639, opened 2026-02-15
Separate position / orientation delta scales for diff IK

brentyi:
#632 uses the same scale for both position and orientation. If you make your robot/task physically bigger, however, you may want to change the position delta scale but not the orientation delta scale.

### DIFF
--- a/src/mjlab/envs/mdp/actions/differential_ik.py
+++ b/src/mjlab/envs/mdp/actions/differential_ik.py
@@ -50,11 +50,18 @@ class DifferentialIKActionCfg(ActionTermCfg):
   """If True, actions are deltas applied to the current frame pose.
   If False, actions are absolute targets."""
 
-  scale: float = 1.0
-  """Scaling factor applied to actions in relative mode.
+  delta_pos_scale: float = 1.0
+  """Scaling factor for position components in relative mode.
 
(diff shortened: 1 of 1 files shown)


kevinzakka (approved):

saikishor:
I'm curious to know if it makes sense to have the position and orientation also support a tuple of 3, that way we can also have a kind of cartesian impedance like behaviour.

What's your opinion on this?

brentyi:
I think adding 3-tuple to the union is reasonable. It's a bit of a slippery slope of course, we could also support general 3x3 transforms to make it so the scales don't have to be axis-aligned. In principle a user might want any `R^n -> R^6` function for converting between raw policy outputs and diff IK deltas, for example to apply a tanh, I don't feel strongly about where to draw the line.

saikishor:
> I think adding 3-tuple to the union is reasonable. It's a bit of a slippery slope of course, we could also support general 3x3 transforms to make it so the scales don't have to be axis-aligned. In principle a user might want any `R^n -> R^6` function for converting between raw policy outputs and diff IK deltas, for example to apply a tanh, I don't feel strongly about where to draw the line.


I agree, it is hard to draw a line. Let's keep it as it is, if there is a need, it can be done

PR #642, opened 2026-02-15
Height scan sensor out of range behaviour

eufrizz:
When training a locomotion policy for a discrete environment with out-of-range raycast readings (pictured), I made a couple of changes to aid training.
- For an out of range ray, instead of returning 0 (which intuitively means an object directly in front of of the sensor), return **maximum sensor distance**. This seems more consistent with what a real depth sensor might show, and preserves physical consistency - as the ground gets further away, the reading increases up to its maximum, instead of jumping back to 0.
- On the Unitree G1 velocity tracking task
  - The height scan reading was clipped from (-1,1), truncating readings more than a metre away - sensor max_distance should determine this instead
  - I normalised the readings in the range [0, 1] by dividing by max_distance for stabler training
  - Reduced max_distance to 3m to increase resolution - the humanoid will probably have a hard time dealing with any drop over 3m anyway.

Feel free to take any or all of these tweaks as you see fit! Can update the changelog with whatever changes are agreed upon.

Screenshot 2026-02-15 at 20 46 33

### DIFF
--- a/src/mjlab/envs/mdp/observations.py
+++ b/src/mjlab/envs/mdp/observations.py
@@ -122,4 +122,10 @@ def height_scan(
     Tensor of shape [B, N] where B is num_envs and N is num_rays.
   """
   sensor: RayCastSensor = env.scene[sensor_name]
-  return sensor.data.pos_w[:, 2].unsqueeze(1) - sensor.data.hit_pos_w[..., 2] - offset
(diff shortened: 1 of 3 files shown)


kevinzakka:
Hi @eufrizz, have a bunch of changes in the perceptive-locomotion branch I'm going to merge sometime soon. It contains a bunch of similar findings.

eufrizz:
Ah sounds good @kevinzakka, will close this PR!

kevinzakka:
I'll keep your PR open just in case :)

PR #646, opened 2026-02-16
Fix height_scan returning ~0 for missed rays

kevinzakka:
Closes #642

- Fix `height_scan` returning ~0 for missed rays; defaults to `max_distance` instead, with a configurable `miss_value` parameter.
- Replace `clip=(-1, 1)` with `scale=1/max_distance` normalization in velocity task config to preserve full sensor range.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -15,6 +15,10 @@ Changed
 Fixed
 ^^^^^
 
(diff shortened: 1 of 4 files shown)


kevinzakka:
@eufrizz decided to percolate the change sooner, thanks again for the initial PR.

PR #649, opened 2026-02-18
Revert video recording from mediapy back to moviepy

kevinzakka:
## Summary

- Reverts video recording from `mediapy` back to `moviepy`. `moviepy` bundles `ffmpeg` via `imageio-ffmpeg`, removing the need for a system `ffmpeg` install.

Addresses feedback in #637.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -8,6 +8,9 @@ Upcoming version (not yet released)
 Changed
 ^^^^^^^
 
+- Reverted video recording from ``mediapy`` back to ``moviepy``. ``moviepy``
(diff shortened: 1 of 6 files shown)


kevinzakka:
cc @rdeits-bd

rdeits-bd:
This would solve our immediate issue, but there is another option -- I think you can do:

```
import imageio_ffmpeg
import mediapy

def tell_mediapy_to_use_imageio_ffmpeg():
    mediapy.set_ffmpeg(imageio_ffmpeg.get_ffmpeg_exe())
```

(you could probably do this at your top level `__init__.py` if you want; I just didn't in this example).

kevinzakka:
@rdeits-bd Thanks! I'll make a new PR with that suggestion.

PR #659, opened 2026-02-20
Read angular velocity directly from cvel

### LINKED ISSUE #658: Potential inefficiency in `Entity.body_link_ang_vel_w`
To compute the angular velocity of a (non root) link, expressed in the world frame, we call:

https://github.com/mujocolab/mjlab/blob/bb2bc159985f8f2c4c2e2d176641d2ac65a448a2/src/mjlab/entity/data.py#L466-L469

which constructs the full linear and angular velocity tensor:

https://github.com/mujocolab/mjlab/blob/bb2bc159985f8f2c4c2e2d176641d2ac65a448a2/src/mjlab/entity/data.py#L286-L293

which grabs the cvel, then does more math to compute the linear velocity from that:

https://github.com/mujocolab/mjlab/blob/bb2bc159985f8f2c4c2e2d176641d2ac65a448a2/src/mjlab/entity/data.py#L20-L31

which we then immediately throw away, since we're just returning the angular component. 

As far as I understand it, the angular part of the cvel *is* the body link angular velocity in the world. Could we return that immediately in `body_link_ang_vel_w` and skip all the other steps?

kevinzakka:
All `*_ang_vel_w` properties (root, body, geom, site) previously computed the full 6-DOF velocity, including a cross-product for the linear part, then discarded everything except the angular component.

Since `cvel[..., 0:3]` is already the angular velocity in world frame (the offset correction only affects the linear part), we can read it directly and skip the unnecessary computation.

Closes #658

### DIFF
--- a/src/mjlab/entity/data.py
+++ b/src/mjlab/entity/data.py
@@ -426,7 +426,7 @@ class EntityData:
   @property
   def root_link_ang_vel_w(self) -> torch.Tensor:
     """Root link angular velocity in world frame. Shape (num_envs, 3)."""
-    return self.root_link_vel_w[:, 3:6]
+    return self.data.cvel[:, self.indexing.root_body_id, 0:3]
 
   @property
   def root_com_pos_w(self) -> torch.Tensor:
@@ -446,7 +446,8 @@ class EntityData:
   @property
   def root_com_ang_vel_w(self) -> torch.Tensor:
     """Root COM angular velocity in world frame. Shape (num_envs, 3)."""
-    return self.root_com_vel_w[:, 3:6]
+    # Angular velocity is the same for link and COM frames.
+    return self.data.cvel[:, self.indexing.root_body_id, 0:3]
 
   @property
   def body_link_pos_w(self) -> torch.Tensor:
@@ -466,7 +467,7 @@ class EntityData:
   @property
   def body_link_ang_vel_w(self) -> torch.Tensor:
     """Body link angular velocities in world frame. Shape (num_envs, num_bodies, 3)."""
-    return self.body_link_vel_w[..., 3:6]
+    return self.data.cvel[:, self.indexing.body_ids, 0:3]
 
   @property
   def body_com_pos_w(self) -> torch.Tensor:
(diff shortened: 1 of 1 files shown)


kevinzakka:
cc @rdeits-bd

src/mjlab/entity/data.py
```diff
@@ -466,7 +467,7 @@ def body_link_lin_vel_w(self) -> torch.Tensor:
   @property
   def body_link_ang_vel_w(self) -> torch.Tensor:
     """Body link angular velocities in world frame. Shape (num_envs, num_bodies, 3)."""
-    return self.body_link_vel_w[..., 3:6]
+    return self.data.cvel[:, self.indexing.body_ids, 0:3]
```
rdeits-bd, line 470: Note that the benefit here is particularly large -- previously this code did the cvel -> vel conversion for *all* bodies any time you asked for the angular velocity of *any* body. 

I think there's still probably some efficiency to be gained here in the future: `self.indexing.body_ids` is a tensor, so I believe this indexing operation will return a copy of the entire `cvel` for all bodies. Typical usage of this property is something like:

```
head_ang_vel_w = entity.data.body_link_ang_vel_w[:, head_body_id, :]
```

which computes that copy of all `cvel`s, then throws away all but one entry. 

I'm not sure what the right way to avoid that is.

kevinzakka: The entity data properties are designed to return quantities for all bodies at once, which pairs naturally with `SceneEntityCfg` for filtering to specific bodies in the reward and observation functions. So the copy from fancy indexing is expected in that workflow.

For the single body case you described, users can already avoid this by using a `frameangvel` builtin sensor attached to the specific body of interest:

```python
BuiltinSensorCfg(
  name="head_ang_vel",
  sensor_type="frameangvel",
  obj=ObjRef(type="xbody", name="head", entity="robot"),
)
```

This reads directly from `sensordata` with no intermediate copy, so it avoids the all bodies indexing entirely. The `xbody` object type here corresponds to the link frame, which is what `body_link_ang_vel_w` returns. For angular velocity specifically both `xbody` and `body` (COM frame) give the same result since angular velocity is the same everywhere on a rigid body, but `xbody` is the correct match for consistency.

rdeits-bd: Oh nice, that's a good idea

kevinzakka: FYI, longer term, I've been looking at adding a **bind** API to mjlab similar to what MuJoCo/MJX provide. Look at [these tests](https://github.com/google-deepmind/mujoco/blob/0f36b0fc2e6a878629452d98ad4abce263647e99/mjx/mujoco/mjx/_src/support_test.py#L160) to get a feel for the API.

kevinzakka: Also final FYI, this doesn't makes a huge difference in FPS for me in the RL loop (133k -> 135k env iterations per second on a 5090). Too much overhead from other parts of the manager-based API.

PR #663, opened 2026-02-22
Overhaul domain randomization with typed `dr` module

kevinzakka:
Domain randomization is one of the most important tools for sim-to-real transfer, yet our previous API was a single `randomize_field` function that treated every model field the same way. This PR replaces it with a typed `dr` module where each function knows which model fields it touches, what derived quantities need recomputation, and what safety invariants to maintain.

### Motivation

The old `randomize_field` had three problems:

1. **No recomputation awareness.** Changing `body_mass` invalidates `body_subtreemass`, `body_invweight0`, and the mass matrix. The user had to know this and call `set_const` manually, otherwise derived quantities would go stale.

2. **No safety invariants.** Changing `geom_size` without updating `geom_rbound` and `geom_aabb` leaves the broadphase with stale bounding volumes, causing missed collisions.

3. **No type structure.** Every call went through the same generic function with a `field` string parameter. Field-specific logic (quaternion composition, bounds recomputation, pseudo-inertia decomposition) could not be attached without special-casing inside the monolith.

### Architecture

Each `dr.*` function is decorated with `@requires_model_fields`, which declares:

- **Which model fields** need per-world storage (expanded from shared to per-env memory)
- **Which recomputation level** to trigger after modification (`set_const`, `set_const_0`, or none)

The `EventManager` collects these declarations at startup, expands the necessary fields once, and after firing all DR terms in a single `apply()` call, triggers the strongest recomputation level needed, exactly once rather than per-term:

```python
from mjlab.envs.mdp import dr

foot_friction = EventTermCfg(
    mode="reset",
    func=dr.geom_friction,
    params={
        "asset_cfg": SceneEntityCfg("robot", geom_names=[".*_foot.*"]),
        "ranges": (0.3, 1.2),
        "operation": "abs",
    },
)
```

### Supported fields

| Category | Functions | Notes |
|----------|-----------|-------|
| **Geom** | `geom_friction`, `geom_pos`, `geom_quat`, `geom_rgba`, `geom_size` | `geom_size` recomputes broadphase bounds automatically |
| **Body** | `body_mass`, `body_com_offset` / `body_ipos`, `body_pos`, `body_quat`, `pseudo_inertia` | `pseudo_inertia` jointly randomizes mass, CoM, and inertia tensor |
| **Joint** | `joint_damping`, `joint_armature`, `joint_friction`, `joint_stiffness`, `joint_limits`, `joint_default_pos`, `encoder_bias` | Aliases map to MuJoCo names (e.g. `joint_damping` → `dof_damping`) |
| **Site** | `site_pos`, `site_quat` | |
| **Camera** | `cam_fovy`, `cam_pos`, `cam_quat`, `cam_intrinsic` | New, for sim-to-real camera variation in vision policies |
| **Light** | `light_pos`, `light_dir` | New, for lighting variation |
| **Tendon** | `tendon_damping`, `tendon_stiffness`, `tendon_friction`, `tendon_armature` | |
| **Actuator** | `pd_gains`, `effort_limits`, `sync_actuator_delays` | Entity-level functions that touch multiple model fields |

### Why mjlab can randomize fields that C MuJoCo cannot

Several fields marked **unsafe** in C MuJoCo are safe to randomize in mjlab due to two architectural differences in MuJoCo Warp:

**No collision BVH.** C MuJoCo builds a static bounding-volume hierarchy for midphase collision pruning. Moving a static body invalidates that tree, and there is no public API to rebuild it. MuJoCo Warp uses NxN or sweep-and-prune broadphase instead, so there is no static tree to invalidate. `body_pos`/`body_quat` are safe for all bodies, including static ones.

**Local bounding boxes.** In MuJoCo Warp, `geom_aabb` is a *local* bounding box (center + half-size in the geom frame). The broadphase transforms it to world space every step using `geom_xpos`/`geom_xmat` from forward kinematics. C MuJoCo caches world-space bounds in the BVH, so changes to `geom_pos`/`geom_quat` make them stale. In MuJoCo Warp, FK updates world-space positions every step, so the local AABB remains valid.

| Field | C MuJoCo | mjlab / MuJoCo Warp | Why |
|-------|----------|---------------------|-----|
| `body_pos`, `body_quat` | Unsafe for static bodies (BVH) | **Safe** | No collision BVH |
| `geom_pos`, `geom_quat` | Unsafe | **Safe** | FK recomputes world positions; local AABB stays valid |
| `geom_size` | Unsafe | **Safe** | `dr.geom_size` recomputes `geom_rbound` and `geom_aabb` inline |

> **FK skip for bare worldbody geoms:** MuJoCo Warp's FK does not update `geom_xpos`/`geom_xmat` for geoms that are both world-welded and not descended from a mocap body (e.g. a ground plane defined directly in XML). This does not affect `dr.*` functions, which target named entities, and all entities are mocap-wrapped automatically. It only matters if you write custom code that directly modifies `sim.model.geom_pos` for a bare worldbody geom.

### Pseudo-inertia randomization

`dr.pseudo_inertia` jointly varies `body_mass`, `body_ipos`, `body_inertia`, and `body_iquat` while guaranteeing the result is always physically valid (positive-definite inertia tensor, positive mass, triangle inequality satisfied).

The parameterization follows [Rucker & Wensing 2022](https://ieeexplore.ieee.org/document/9690029): the pseudo-inertia matrix $J \succ 0$ is factored via Cholesky as $J = LL^\top$, then perturbed by an upper-triangular matrix $U$ so that $J' = (UL)(UL)^\top$. The result is decomposed back to MuJoCo fields via eigendecomposition, which is exact for any perturbation magnitude.

The 10 parameters of $U$ control distinct physical effects:
- **`alpha`**: global density scale (mass and inertia scale by $e^{2\alpha}$, CoM unchanged)
- **`d1, d2, d3`**: axis-aligned stretch/compress
- **`s12, s13, s23`**: shear (redistributes mass asymmetrically)
- **`t1, t2, t3`**: CoM shift along body-frame axes

```python
body_inertia_dr = EventTermCfg(
    mode="reset",
    func=dr.pseudo_inertia,
    params={
        "asset_cfg": SceneEntityCfg("robot", body_names=["link_2", "link_3"]),
        "alpha_range": (-0.5, 0.5),
    },
)
```

yam_inertia_rand_teaser

*`dr.pseudo_inertia` with `alpha_range=(-0.5, 0.5)` on links 2 and 3. The inertia ellipsoids resize each episode reset while other links remain unchanged.*

> `dr.body_mass` still exists but emits a warning: scaling mass without scaling inertia is only correct for a point mass at the CoM (zero rotational inertia). For the common DR use case of density uncertainty, use `pseudo_inertia` with `alpha_range`.

### Viewer integration

The native viewer syncs per-world DR model fields from the GPU before each render. MuJoCo's built-in visualization toggles work correctly against the randomized model:

- Geom appearance (`geom_rgba`, `geom_size`, `geom_pos`, `geom_quat`)
- Inertia ellipsoids (press `I`) reflect randomized `body_mass`, `body_inertia`, `body_iquat`
- Camera frustums (press `Q`) reflect randomized `cam_fovy`, `cam_intrinsic`
- Light positions and directions

yam_combined_rand_teaser

*Cube color (`dr.geom_rgba`), cube size (`dr.geom_size`), and link 2/3 orientations (`dr.body_quat`) randomized each episode reset. Broadphase bounds are recomputed automatically after size changes.*

### TerrainEntity

Terrain is now a proper `Entity` subclass (`TerrainEntity`), which means DR functions can target terrain parameters via `SceneEntityCfg("terrain", ...")`, for example randomizing terrain friction or terrain-mounted camera/light positions. `TerrainImporter` / `TerrainImporterCfg` remain as aliases.

### Breaking changes

- **`randomize_field` is removed.** Replace with the corresponding typed `dr.*` function (e.g. `dr.geom_friction`, `dr.body_mass`).
- **`EventTermCfg.domain_randomization` is removed.** The `@requires_model_fields` decorator handles field tracking automatically.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -53,6 +53,11 @@ Added
   correctly when the corresponding fields are randomized. See
   :doc:`randomization` for viewer-specific caveats.
 
+- ``DebugVisualizer`` now supports ellipsoid visualization via
+  ``add_ellipsoid``.
+
+- Visualizers display the realtime factor alongside FPS.
+
 - Terrain is now a proper ``Entity`` subclass (``TerrainEntity``). This
   allows domain randomization functions to target terrain parameters
   (friction, cameras, lights) via ``SceneEntityCfg("terrain", ...)``.
--- a/src/mjlab/terrains/__init__.py
+++ b/src/mjlab/terrains/__init__.py
@@ -51,6 +51,7 @@ from mjlab.terrains.terrain_generator import SubTerrainCfg as SubTerrainCfg
 from mjlab.terrains.terrain_generator import TerrainGenerator as TerrainGenerator
 from mjlab.terrains.terrain_generator import TerrainGeneratorCfg as TerrainGeneratorCfg
 
+# TODO(kevin): remove these aliases, see https://github.com/mujocolab/mjlab/issues/667
 # Backwards compatibility aliases (deprecated).
 _DEPRECATED_ALIASES = {
   "TerrainImporter": "TerrainEntity",


kevinzakka:
Tests will pass once https://github.com/google-deepmind/mujoco_warp/pull/1174 lands.

src/mjlab/envs/mdp/dr/body.py
```diff
@@ -0,0 +1,438 @@
...
+  distribution: Distribution = "uniform",
+  asset_cfg: SceneEntityCfg = _DEFAULT_ASSET_CFG,
+) -> None:
+  r"""Physics-consistent inertial randomization via the pseudo-inertia matrix.
+
+  Jointly randomizes ``body_mass``, ``body_ipos``, ``body_inertia``, and ``body_iquat``
+  while guaranteeing exact physical consistency for any perturbation magnitude.
+  Triggers ``set_const`` recomputation.
+
+  The parameterization follows https://ieeexplore.ieee.org/document/9690029
```
rdeits-bd: FYI, non-paywalled link to the same article is https://par.nsf.gov/servlets/purl/10347458

src/mjlab/envs/mdp/dr/_core.py
```diff
@@ -0,0 +1,411 @@
...
+) -> None:
+  """Apply the randomization operation."""
+  if operation == "add":
+    model_field[env_grid, entity_grid] = indexed_data + random_values
+  elif operation == "scale":
+    model_field[env_grid, entity_grid] = indexed_data * random_values
+  elif operation == "abs":
+    model_field[env_grid, entity_grid] = random_values
+  else:
+    raise ValueError(f"Unknown operation: {operation}")
```
rdeits-bd: I know the "add", "scale", "abs" string API is inherited from IsaacLab, but this kind of interface always seems unnecessarily limiting to me. Despite the fact that we're working in a totally dynamic language, we only allow users to pick from exactly three operations, and the only way to add more is to monkey-patch the internals of mjlab. Likewise for the distributions -- there are only three built-in options and it's impossible to add more. 

What about something like:

```py
class Operation(ABC):
  @staticmethod
  @abstractmethod
  def apply(model_field, env_grid, entity_grid, indexed_data, random_values):
    ...

  @staticmethod
  @abstractmethod
  def base_values(env, field):
    ....

class add(Operation):
  ...

class abs(Operation):
  ...
```

or, alternatively, a single `Operation` class whose constructor takes callback functions, with `abs`, `add`, etc. as const instances of that class. 

In either case, you'd end up with roughly the same surface syntax, e.g. `"operation": dr.add` instead of `"operation": "add"`. But it would suddenly be easy for users to create their own new behaviors without modifying mjlab. 

The downside is that you drift farther from syntax compatibility with IsaacLab.

src/mjlab/envs/mdp/dr/body.py
```diff
@@ -0,0 +1,438 @@
...
+  env: ManagerBasedRlEnv,
+  env_ids: torch.Tensor | None,
+  ranges: Ranges,
+  asset_cfg: SceneEntityCfg = _DEFAULT_ASSET_CFG,
+  distribution: Distribution = "uniform",
+  operation: Operation = "add",
+  axes: list[int] | None = None,
+  shared_random: bool = False,
+) -> None:
+  """Randomize body COM offset (body_ipos). Triggers ``set_const``."""
```
rdeits-bd, line 236: Could this, and various other similar instances elsewhere, be simplified to:

```py
body_com_offset = _generic_field_randomization("body_ipos", RecomputeLevel.set_const)
```

using something like:

```py
def _generic_field_randomization(field_name, recompute):
  @requires_model_fields(field_name, recompute=recompute):
  def f(<args here>):
    _recompute_model_field(<args here>)
```

? That might reduce the boilerplate needed here. The downside is that attaching the docstring becomes more awkward. 

Or, alternatively, if all of these common args were wrapped up in a dataclass, then I think this would look like:

```py
@requires_model_fields("body_ipos", recompute=RecomputeLevel.set_const)
def body_com_offset(model_field_args = ModelFieldArgs()):
  """<docstring goes here>"""
  _randomize_model_field("body_ipos", model_field_args)
```

which would match the structure you've already got with less boilerplate. Unless you're using different values of the various individual args for these different fields?

src/mjlab/envs/mdp/dr/body.py
kevinzakka: Thanks, updated the docs link!

src/mjlab/envs/mdp/dr/_core.py
kevinzakka: Agreed, implemented this. Operation and Distribution are now dataclasses that take callback functions, with the built-in instances (`dr.abs`, `dr.add`, `dr.scale`, `dr.uniform`, etc.) as module-level constants. Strings still resolve to these for convenience. See `_types.py` and the new **Custom Operation/Distribution** sections in the docs.

src/mjlab/envs/mdp/dr/body.py
kevinzakka: Considered this but decided to keep explicit function definitions. The boilerplate is small after the `_core` refactor, and explicit defs are easier to navigate (grep, IDE jump-to-definition) and attach docstrings to. We can revisit this in the future though!

PR #676, opened 2026-02-24
fix types of default values

jgueldenstein:
When using the default plane config I got the following error:

```
╭─ Invalid input to tyro.cli() ────────────────────────────────────────────────╮
│ The default value of the env.scene.terrain field could not be matched to any │
│ of its subcommands.                                                          │
│ ──────────────────────────────────────────────────────────────────────────── │
│                                                                              │
│ env.scene.terrain:terrain-entity-cfg was not a match because:                │
│ • Field lights has invalid default                                           │
│ • Field lights.0 has invalid default                                         │
│ • Field lights.0.cutoff has invalid default                                  │
│ • Default value 45 with type int does not match type <class 'float'>         │
│                                                                              │
│ env.scene.terrain:none was not a match because:                              │
│ • Default type <class 'mjlab.terrains.terrain_entity.TerrainEntityCfg'> is   │
│   not None                                                                   │
│                                                                              │
│ ──────────────────────────────────────────────────────────────────────────── │
│ Debugging: check that the field default matches a member of the union type.  │
╰──────────────────────────────────────────────────────────────────────────────╯ 
```

The scene was configured as follows:
```python
    scene = SceneCfg(
        entities={"robot": get_x02_robot_cfg()},
        sensors=(feet_contact_sensor_cfg, self_collision_cfg),
        env_spacing=0.0,
        terrain=TerrainEntityCfg(terrain_type="plane"),
    )
```

This PR fixes this problem.

### DIFF
--- a/src/mjlab/utils/spec_config.py
+++ b/src/mjlab/utils/spec_config.py
@@ -271,13 +271,13 @@ class LightCfg(SpecCfg):
   """Light type ("spot" or "directional")."""
   castshadow: bool = True
   """Whether light casts shadows."""
-  pos: tuple[float, float, float] = (0, 0, 0)
+  pos: tuple[float, float, float] = (0.0, 0.0, 0.0)
   """Light position (x, y, z)."""
-  dir: tuple[float, float, float] = (0, 0, -1)
+  dir: tuple[float, float, float] = (0.0, 0.0, -1.0)
   """Light direction vector (x, y, z)."""
-  cutoff: float = 45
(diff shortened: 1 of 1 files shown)


kevinzakka:
Thanks for catching this! I had the same fix in flight and just merged it in #678. Your PR correctly identified the issue. I also found a few more int/float mismatches in `terrain_entity.py`, `sim.py`, and `primitive_terrains.py` that I included there. Closing in favor of #678 but really appreciate you opening this!

PR #692, opened 2026-02-26
Add command line argument to train, play and evaluate script to specify checkpoint name for wandb stored runs

jgueldenstein:
Saving runs to wandb is a really great feature.
The changes in this PR allow to add a new argument to the play, train and evaluate script to specify which model is loaded (rather than the newest one).

Usage is:
`uv run play <task-name> --wandb-run-path entity/project/run-id --wandb-checkpoint-name model_4000.pt`

If not specified, the previous behavior (loading the newest run) applies.

### DIFF
--- a/src/mjlab/scripts/play.py
+++ b/src/mjlab/scripts/play.py
@@ -25,6 +25,7 @@ class PlayConfig:
   registry_name: str | None = None
   wandb_run_path: str | None = None
   wandb_checkpoint_name: str | None = None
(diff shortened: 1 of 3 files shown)


kevinzakka (approved):
Thanks @jgueldenstein!

I made a small fix: the `checkpoint_name` was being wrapped in `Path()` unnecessarily (it should stay a str to match the other branch), and added validation so you get a clear error if the requested checkpoint doesn't exist in the run.

PR #693, opened 2026-02-26
Fix viewer FPS drop caused by physics stepping loop

oxkitsune:
The physics/render decoupling in #663 introduced a while loop that drains the full sim-time budget in a single tick() call. At 1x speed this means multiple `env.step()` calls back-to-back per frame, starving the renderer and dropping FPS from ~30 to ~4 (on my M3 macbook at least, on a 5090 it was around 20-25!).

I fixed this by replacing the while loop with a single if statement, so at most one `env.step()` runs per `tick()` call. 
The catch-up still works, because when the budget exceeds `step_dt`, every subsequent `tick()` (~1ms apart in the `run()` loop) steps immediately until the budget is drained. 

The difference is that each step yields back to `run()`, giving the renderer a chance to check `time_until_next_render` between steps, instead of holding the entire frame hostage.

Before|After
--|--
before|after

### DIFF
--- a/src/mjlab/viewer/base.py
+++ b/src/mjlab/viewer/base.py
@@ -247,16 +247,17 @@ class BaseViewer(ABC):
 
     wall_dt = self._timer.tick()
 
-    # Step physics based on accumulated sim-time budget.
+    # Step physics based on accumulated sim-time budget.  At most one
+    # env.step() per tick so that physics never starves the renderer.
+    # The budget is allowed to accumulate across ticks for catch-up, but
+    # is capped to avoid a spiral of death after long stalls.
     if not self._is_paused:
       step_dt = self.env.unwrapped.step_dt
       self._sim_time_budget += wall_dt * self._time_multiplier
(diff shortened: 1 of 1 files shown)


kevinzakka:
Thanks for catching this @oxkitsune! You nailed the root cause.

We went with a slightly different approach in #694 that keeps the `while` loop but makes it self-limiting with a two-clock design (tracked vs actual sim time, inspired by dm_control's viewer). The extra insight was that `wall_dt` includes the previous tick's step execution time, creating a feedback spiral. We break it by subtracting step wall time from the budget.

Your `if` approach works great at 1x but gets fragile at other speed multipliers since it depends on the 1ms sleep cadence in `run()`. The two-clock version handles all speeds correctly.

Closing in favor of #694. Thanks again for the diagnosis and the before/after screenshots!

oxkitsune:
Thanks for fixing!

kevinzakka:
Thank you @oxkitsune!

PR #700, opened 2026-02-27
Revamp Viser term plotter

louislelay:
Reworked the rewards/metrics plotter for the Viser viewer:

- Terms are grouped by prefix (e.g. `track_lin_vel`, `track_ang_vel` → `track` folder) with All/None bulk toggle
- Combined overlay plot to compare enabled terms side by side, each with its own color
- Individual plots per enabled term, lazily created/removed
- Removed the `max_terms` cap

https://github.com/user-attachments/assets/62c901c6-a415-4c9b-bba9-6bf6784ee4df

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ uv run train Mjlab-Velocity-Flat-Unitree-G1 --env.scene.num-envs 4096
 
 ```bash
 uv run train Mjlab-Velocity-Flat-Unitree-G1 \
-  --gpu-ids 0 1 \
+  --gpu-ids "[0, 1]" \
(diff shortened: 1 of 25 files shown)


louislelay:
Removed the overlay on advice from @saikishor, it was a bit overkill.

image

brentyi (commented):
Cool!

Only nit from me: do we need the "Reward: " prefix in all of the labels? For example, "Reward: foot_slip" could just be "foot_slip", "Reward: Select terms" could just be "Select terms", etc.

louislelay:
Thanks @brentyi, just did!

image

PR #701, opened 2026-02-27
Add BeyondMimic to the Publications section

louislelay:


kevinzakka:
BeyondMimic original release results were not obtained with mjlab.

louislelay:
I know but as in the last version published on arXiv, they cite mjlab i thought it was relevant!

PR #709, opened 2026-03-01
xml actuator target match

omarrayyann:
https://github.com/mujocolab/mjlab/blob/177cede03ff43d082e283decd25ee66fc039b096/src/mjlab/actuator/xml_actuator.py#L60-L61

`_find_actuator_for_target` doesnt match with the target if is is inside an `<attach prefix="...">`. Spec stores full prefixed name but `Entity.find_tendon` strip prefixes and return local names. 

the PR fix is not ideal btw since it's ambiguous if sub-models share a similar local target name.


kevinzakka:
@omarrayyann Do you have a MRE I can test locally so I can come up with best possible fix?

omarrayyann:
Resolved by #714

PR #714, opened 2026-03-02
Fix actuator target resolution for entities with internal attach prefixes

kevinzakka:
When an entity's `spec_fn` uses `MjSpec.attach(prefix=...)` internally (e.g., composing a robot from sub-assemblies), all actuator types failed to resolve their targets:

- **Builtin actuators** hit a MuJoCo compile error (`unknown transmission target 'elbow'`) because `spec.add_actuator(target="elbow")` couldn't find the joint actually named `"arm/elbow"` in the spec.
- **XML actuators** found no matching actuators because `_find_actuator_for_target` compared `actuator.target` (`"arm/elbow"`) against the stripped name (`"elbow"`).

**Root cause:** `_add_actuators` passed stripped names (from `find_joints` etc.) to `edit_spec`, but MuJoCo's spec APIs need the full names including any internal attach prefixes.

**Fix:** `_add_actuators` now resolves full spec names from the element references via `target_ids` and passes those to `edit_spec`. User-facing names (`actuator.target_names`, `entity.joint_names`) remain stripped.

Closes #709.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -119,6 +119,9 @@ Changed
 Fixed
 ^^^^^
 
+- Fixed actuator target resolution for entities whose ``spec_fn`` uses
+  internal ``MjSpec.attach(prefix=...)``
+  (`#709 <https://github.com/mujocolab/mjlab/issues/709>`_).
 - Fixed viewer physics loop starving the renderer by replacing the single
   sim-time budget with a two-clock design (tracked vs actual sim time).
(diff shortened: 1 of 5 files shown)


kevinzakka:
cc @omarrayyann

omarrayyann:
Thanks! will test it out now and let you know.

omarrayyann:
@kevinzakka works fine now! thank you

PR #721, opened 2026-03-03
Add export-scene CLI and fix scene XML export

kevinzakka:
`Scene.to_zip` relied on `MjSpec.to_zip` which produced XML with broken asset paths and could not serialize buffer textures. This replaces it with `Scene.write`, which emits clean XML with correct meshdir references and copies mesh assets into an `assets/` subdirectory.

New xml utilities (`fix_spec_xml`, `strip_buffer_textures`) handle known `MjSpec.to_xml` limitations: empty default elements, duplicate nested defaults, and buffer textures that cannot round trip through XML. `Entity.write_xml` also applies these fixups. `Scene.to_zip` is kept as a deprecated wrapper.

New `export-scene` debug script for inspecting and sharing scenes:

```
export-scene g1 --output-dir /tmp/g1
export-scene Mjlab-Velocity-Flat-Unitree-Go1 --output-dir /tmp/task
export-scene yam --output-dir /tmp/yam --zip True
export-scene badname  (prints available targets)
```

Accepts a positional target that resolves against the task registry or a hardcoded entity lookup (`g1`, `go1`, `yam`). Output directory is cleaned before each export to avoid mixing stale assets.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -81,6 +81,7 @@ Table of Contents
    source/training/distributed_training
    source/training/cloud
    source/debugging/nan_guard
+   source/debugging/export_scene
 
 .. toctree::
    :maxdepth: 2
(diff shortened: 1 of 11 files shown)


kevinzakka:
cc @erikfrey

erikfrey:
I love it!  We will eventually fix spec to_zip, in the meantime this makes sense.  I’ll ping you when spec to_zip is working better.

saikishor:
Thank you. I badly wanted this too

PR #730, opened 2026-03-03
Make neighbor env count configurable in ViewerConfig

kevinzakka:
Replace the hardcoded `_MAX_EXTRA_ENVS` constant with a `ViewerConfig.max_extra_envs` field, letting users control how many neighboring environments are rendered around the tracked subject. Defaults to 2, preserving existing behavior.


rdeits-bd (approved):
Thanks! We've been modifying `_MAX_EXTRA_ENVS` locally, but this is much nicer.

PR #735, opened 2026-03-05
Add apply_body_impulse event and force visualization

kevinzakka:
Adds transient external force perturbations to bodies via a new `apply_body_impulse` step event, with configurable duration, cooldown, and application point offset.

https://github.com/user-attachments/assets/5413b218-7e9c-4f86-9843-589f29334867

*Random impulses hit the punching bag while mouse perturbation is applied simultaneously. The two force sources coexist because programmatic forces flow through `xfrc_applied` and mouse forces through `qfrc_applied`.*

- `apply_body_impulse` with `VizCfg` for arrow rendering. Events can now implement `debug_vis()`, delegated through `EventManager`.
- Native viewer syncs `xfrc_applied` GPU to CPU for rendering and converts mouse perturbation to `qfrc_applied` via `mj_applyFT` so the two never overwrite each other.
- Punching bag demo (`scripts/demos/body_impulse.py`).

### DIFF
--- a/src/mjlab/envs/mdp/events.py
+++ b/src/mjlab/envs/mdp/events.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 import torch
@@ -9,6 +10,7 @@ import torch
 from mjlab.entity import Entity
 from mjlab.managers.scene_entity_config import SceneEntityCfg
 from mjlab.utils.lab_api.math import (
+  quat_apply,
   quat_from_euler_xyz,
   quat_mul,
   sample_uniform,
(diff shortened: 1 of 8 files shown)


bd-aadubredu:
thank you! LGTM!

src/mjlab/envs/mdp/events.py
```diff
@@ -351,3 +354,225 @@ def apply_external_force_torque(
...
+
+  def reset(self, env_ids: torch.Tensor | slice | None = None) -> None:
+    if env_ids is None:
+      env_ids = slice(None)
+
+    # Clear forces for reset envs.
+    if isinstance(env_ids, slice):
+      reset_ids = env_ids
+    else:
+      reset_ids = env_ids
```
rdeits-bd, line 560: FYI this `if` statement is tautological -- both branches do the same thing

PR #742, opened 2026-03-06
Add mean action acceleration metric & clean up velocity env

louislelay:
In PR #700 while addressing Brent's comment, I pushed a minimal metrics implementation to the velocity tracking env that I was using to check for cross-contamination between metrics and rewards in the plotter. This PR removes that test code.

Since we didn't have a proper example of a metrics implementation, I took the opportunity to add one: a general-purpose mean_action_acc metric in mdp/metrics.py, wired into the velocity env config.


kevinzakka:
Thanks @louislelay!

PR #744, opened 2026-03-06
Change to Base class type for default runner check

ManuelActisCassa:
Right now, can only pass **on policy runner configuration** for mjlab_task (train, play).

Changed that to **base runner cfg** (more generic, all runner configurations will inherit from this one), to allow in the future other type of runners (off policy, fast SAC, ...) without errors


louislelay:
Nice @ManuelActisCassa, congrats on your first PR to mjlab! :)

PR #745, opened 2026-03-06
Sort actuators (optional flag)

etaoxing:
These two changes improve portability when exporting mjlab defined specs to XML.

Add `EntityCfg.sort_actuators` boolean

- defaults to False to preserve original behavior
- when True, sorts actuators before `edit_spec()` call so that `model.ctrl` order will match `model.joint` order
- mjlab custom actuators are added after any XML defined actuators

Set ctrlrange in `create_position_actuator()` for consistency with other actuators
- was not defined originally since `ctrllimited=False`, but this change shouldn't affect https://github.com/mujocolab/mjlab/pull/354

### DIFF
--- a/src/mjlab/entity/entity.py
+++ b/src/mjlab/entity/entity.py
@@ -80,7 +80,11 @@ class EntityCfg:
     default_factory=lambda: (lambda: mujoco.MjSpec())
   )
   articulation: EntityArticulationInfoCfg | None = None
-  sort_actuators: bool = True
+  sort_actuators: bool = False
+  """When True, reorder actuators so that ``model.ctrl`` follows joint/tendon/site
+  definition order rather than the order actuators appear in the config. XML actuators
+  are excluded from sorting and always retain their declaration order.
+  """
 
   # Editors.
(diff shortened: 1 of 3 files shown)


kevinzakka:
@etaoxing Can you add tests? I'm super wary of unintentional side effects from such changes.

etaoxing:
@kevinzakka Added a couple tests!

Yes also concerned about downstream effects, so before kept original behavior with `sort_actuators=False`. 

Latest changes now set `sort_actuators=True` and seems like the test suite passes. Can revert to `sort_actuators=False` if we want to preserve original behavior. 

This PR should not change mjlab behavior though, just the underlying mjspec that mjlab produces.

kevinzakka:
Hey @etaoxing, made a few tweaks and pushed to your branch:

- Changed `sort_actuators` default to False to preserve backward compatibility. Users can opt in when they need it.
- Cleaned up the sorting logic and tests.
- Added comments explaining the `inheritrange`/`ctrllimited` interaction and the `ctrlrange` derivation.

Could you take a look and let me know if you're good with the changes?

etaoxing:
@kevinzakka looks great, still works for me locally

PR #747, opened 2026-03-07
Fuse delayed builtin actuators into batched buffer operations

kevinzakka:
Adds `DelayedBuiltinActuatorGroup` that merges N per-actuator delay buffer append + compute calls into one when actuators share the same delay config, builtin type, and transmission type.

Absorbed actuators retain access to the shared buffer via aliasing, so per-actuator `reset()` and `set_lags()` continue to work transparently.

Benchmarked on G1 (37 DOF, 6 actuator groups, RTX 5090). Unfused delayed actuators add 11-26% overhead depending on env count. Fusing recovers 80-86% of that (e.g. 26% → 5% at 1024 envs, 11% → 2% at 8192 envs).

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -5,6 +5,12 @@ Changelog
 Upcoming version (not yet released)
 -----------------------------------
 
+Added
+^^^^^
+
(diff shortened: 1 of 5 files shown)


bd-aadubredu:
thanks! LGTM

PR #751, opened 2026-03-08
Add W&B run tab to viser viewer

omarrayyann:
- "W&B Run" tab to the vizer play viewer when play is ran with `--wandb_run_path` 
- allow switching/fetching checkpoints from wandb as the run progresses. rather than having to rerun play

Screenshot 2026-03-07 at 3 56 27 PM

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -36,6 +36,9 @@ Added
   once per physics substep inside the decimation loop. The per substep
   values are averaged within each environment step, so episode averages
   remain comparable to regular per step metrics.
+- Added a Checkpoints tab to the Viser play viewer for hot-swapping
+  checkpoints without restarting. Works with local directories and W&B
+  runs (:issue:`751`). Contribution by @omarrayyann.
 
 Changed
 ^^^^^^^


omarrayyann:
@louislelay done

kevinzakka:
Sorry for the delay, will review this weekend!

kevinzakka (approved):
Thanks @omarrayyann, this is awesome!!

I made a few changes: rebased onto main, replaced webbrowser.open() with a clickable HTML link, deduplicated the `runner.load()` pattern, added missing error handling for checkpoint name parsing, and used `threading.Event` for thread-safe UI suppression.

src/mjlab/scripts/play.py
```diff
@@ -194,6 +204,78 @@ def __call__(self, obs) -> torch.Tensor:
...
+    else:
+      import wandb
+
+      api = wandb.Api()
+      run_path = str(cfg.wandb_run_path)
+      wandb_run = api.run(run_path)
+      _log_root = log_root_path  # pyright: ignore[reportPossiblyUnboundVariable]
+
+      def fetch_available_wandb() -> list[tuple[str, str]]:
+        wandb_run.update()
```
chungmin99, line 251: Is there any reason why this needs to be `wandb_run.update()` ? This seems to work also with `.load()` as well, and might be better suited for the use case scenario here (fetching runs). 

(Personally I was running into some issues with httperror / permission issues with `update`)

PR #780, opened 2026-03-11
Add curriculum for reward params

saikishor:
I've a use case where I would like to modulate the reward params with curriculum. I saw there is currently no method to do so, so I've implemented it and also moved the common codebase to envs folder and added some tests.

I'm also curious to know if we need one that handles both weight and params?

Please let me know what you think.

Thank you


kevinzakka:
Thanks for this @saikishor! I liked the idea and ran with it in a slightly different direction. Take a look at #791 and let me know what you think. The main change is merging `reward_weight` and `reward_params` into a single `reward_curriculum` function, and using `>=` instead of `>` for step thresholds.

PR #790, opened 2026-03-12
Fix stepping stones gap around platform and scale size with difficulty

kevinzakka:
`BoxSteppingStonesTerrainCfg` used a fixed margin (`max(avg_s_max, stone_size_variation * 2) / 2`) to skip stones near the platform. This excluded stones that did not actually overlap the platform, creating a visible gap. Stones are now only skipped when their center falls inside the platform. Edges that extend under the platform are allowed since the platform covers them.

Stone base size was also fixed regardless of difficulty. It now interpolates from the large end of `stone_size_range` at difficulty 0 to the small end at difficulty 1, matching how `stone_distance_range` already scales.

stepping_stones_compare

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -30,6 +30,9 @@ Added
 Changed
 ^^^^^^^
 
+- ``BoxSteppingStonesTerrainCfg`` stone size now decreases with difficulty,
+  interpolating from the large end of ``stone_size_range`` at difficulty 0
+  to the small end at difficulty 1 (:issue:`785`).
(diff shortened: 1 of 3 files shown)


kevinzakka:
cc @Wang-Haitong

Wang-Haitong:
Hi @kevinzakka, thanks for the fix. 

However, regarding scaling size with difficulty, I think currently at difficulty 0, both distance and size are the average of min and max, would it be better if they start from min for stone distance, and max for stone size.

PR #791, opened 2026-03-12
Add unified reward_curriculum function

kevinzakka:
Alternative to #780. Builds on @saikishor's idea of consolidating the duplicated reward_weight and adding param support, with a few design changes:

- Single reward_curriculum function instead of separate reward_weight and reward_params. Each stage can carry weight, params, or both.
- Uses >= instead of > for step thresholds, so a step: 0 stage applies on the first call.
- Plain dict.update instead of deep merge for params. Simpler and more predictable.
- Adds a "Reward curriculum" section to the docs with usage examples.

Closes #780.

### DIFF
--- a/src/mjlab/envs/mdp/curriculums.py
+++ b/src/mjlab/envs/mdp/curriculums.py
@@ -53,6 +53,13 @@ def reward_curriculum(
       if "weight" in stage:
         reward_term_cfg.weight = stage["weight"]
       if "params" in stage:
+        unknown = stage["params"].keys() - reward_term_cfg.params.keys()
+        if unknown:
+          raise KeyError(
+            f"reward_curriculum: stage at step {stage['step']} sets"
+            f" unknown param(s) {unknown} on reward term"
+            f" '{reward_name}'. Check for typos."
+          )
         reward_term_cfg.params.update(stage["params"])
   result: dict[str, torch.Tensor] = {
     "weight": torch.tensor(reward_term_cfg.weight),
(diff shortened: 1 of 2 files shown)


src/mjlab/envs/mdp/curriculums.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+  """
+  del env_ids  # Unused.
+  reward_term_cfg = env.reward_manager.get_term_cfg(reward_name)
+  for stage in stages:
+    if env.common_step_counter >= stage["step"]:
+      if "weight" in stage:
+        reward_term_cfg.weight = stage["weight"]
+      if "params" in stage:
+        reward_term_cfg.params.update(stage["params"])
```
saikishor, line 63: @kevinzakka 
what do you think about adding a check to verify that the number of keys before and after is the same?

If the user sets "max_velocity" instead of "max_vel", it would update and silently fail right?. It's just to deal with typos somehow and not to inroduce new unknown parameters

src/mjlab/envs/mdp/curriculums.py
```diff
@@ -0,0 +1,65 @@
...
+          {"step": 12000, "weight": -0.1},
+          {"step": 24000, "weight": -1.0, "params": {"max_vel": 1.0}},
+        ],
+      },
+    )
+  """
+  del env_ids  # Unused.
+  reward_term_cfg = env.reward_manager.get_term_cfg(reward_name)
+  for stage in stages:
+    if env.common_step_counter >= stage["step"]:
```
saikishor, line 52: Very good catch

src/mjlab/envs/mdp/curriculums.py
kevinzakka: Great idea, done!

kevinzakka:
Thanks for the review @saikishor!

PR #794, opened 2026-03-13
Fix write_root_com_velocity_to_sim with tensor env_ids

### LINKED ISSUE #793: `write_root_com_velocity_to_sim()` fails with tensor `env_ids`
## Summary

Calling `write_root_com_velocity_to_sim()` on a floating-base G1 can fail with:

```text
RuntimeError: linalg.cross: inputs must have the same number of dimensions.
```

The attached repro shows that:

- `--method api` fails when calling `robot.write_root_com_velocity_to_sim(...)`
- `--method manual` succeeds when converting COM velocity to root-link velocity first and then calling `write_root_link_velocity_to_sim(...)`

## Environment

- `mjlab==1.2.0`
- `mujoco-warp==3.5.0.2`
- `torch==2.7.0`
- Python `3.12`

## Minimal Repro

Repro repository:

- https://github.com/EGalahad/mjlab_write_root_com_vel

Files:

- `repro.py`
- `pyproject.toml`

Run:

```bash
uv run repro.py --method api    # fails with shape error
uv run repro.py --method manual # succeeds
```

## Actual Behavior

`uv run repro.py --method api` fails with:

```text
Traceback (most recent call last):
  File ".../repro.py", line ..., in main
    robot.write_root_com_velocity_to_sim(root_com_velocity, env_ids=env_ids)
  File ".../mjlab/entity/entity.py", line 836, in write_root_com_velocity_to_sim
    self._data.write_root_com_velocity(root_velocity, env_ids)
  File ".../mjlab/entity/data.py", line 140, in write_root_com_velocity
    lin_vel_link = lin_vel_com - torch.cross(ang_vel_w, com_offset_w, dim=-1)
RuntimeError: linalg.cross: inputs must have the same number of dimensions.
```

## Expected Behavior

`write_root_com_velocity_to_sim()` should accept tensor `env_ids` and set the root COM velocity without a shape error.

## Root Cause

In `mjlab/entity/data.py`:

```python
env_ids = self._resolve_env_ids(env_ids)
com_offset_b = self.model.body_ipos[:, self.indexing.root_body_id]
quat_w = self.data.qpos[env_ids, self.indexing.free_joint_q_adr[3:7]]
com_offset_w = quat_apply(quat_w, com_offset_b[env_ids])
```

When `env_ids` is a tensor, `_resolve_env_ids()` converts it to shape `(N, 1)`. Then:

- `ang_vel_w` has shape `(N, 3)`
- `com_offset_b[env_ids]` becomes `(N, 1, 3)`
- `com_offset_w` becomes `(N, 1, 3)`

This causes `torch.cross(ang_vel_w, com_offset_w, dim=-1)` to fail because the operands do not have the same number of dimensions.

The core issue is that `env_ids` should not be reused in its `(N, 1)` form when indexing `com_offset_b`. It should be squeezed back to `(N,)` first.

## Minimal Fix

The minimal fix is to keep the current `qpos` indexing, but squeeze `env_ids` before indexing `com_offset_b`:

```python
env_ids = self._resolve_env_ids(env_ids)
env_ids_flat = env_ids.squeeze(-1) if isinstance(env_ids, torch.Tensor) else env_ids

com_offset_b = self.model.body_ipos[:, self.indexing.root_body_id]
quat_w = self.data.qpos[env_ids, self.indexing.free_joint_q_adr[3:7]]
com_offset_w = quat_apply(quat_w, com_offset_b[env_ids_flat])
```

This keeps:

- `quat_w` at shape `(N, 4)`
- `com_offset_b[env_ids_flat]` at shape `(N, 3)`
- `com_offset_w` at shape `(N, 3)`

so `torch.cross(ang_vel_w, com_offset_w, dim=-1)` works as expected.

kevinzakka:
`write_root_com_velocity` called `_resolve_env_ids`, which reshapes tensor env_ids to `(N, 1)` for 2D advanced indexing (e.g. `qpos[env_ids, col_indices]`). But it also used the reshaped ids for 1D indexing into `com_offset_b`, producing shape `(N, 1, 3)` instead of `(N, 3)` and causing a dimension mismatch in `torch.cross`. The reshaped ids were also passed to `write_root_velocity`, which called `_resolve_env_ids` again, creating `(N, 1, 1)`.

The fix avoids `_resolve_env_ids` entirely in this method. Instead it indexes `qpos` in two steps (`qpos[:, cols][env_ids]`) so all tensors use consistent 1D indexing, and passes the original `env_ids` through to `write_root_velocity`.

Fixes #793

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -56,6 +56,8 @@ Fixed
 - ``create_velocity_actuator`` no longer sets ``ctrllimited=True`` with
   ``inheritrange=1.0``. This caused a ``ValueError`` for continuous joints
   (e.g. wheels) that have no position range defined (:issue:`787`).
+- ``write_root_com_velocity_to_sim`` no longer fails with tensor ``env_ids``
+  on floating base entities (:issue:`793`).
 - Joint limits for unlimited joints are now set to [-inf, inf] instead of
   [0, 0]. Previously the zero range caused incorrect clamping for entities
   with unlimited hinge or slide joints.
--- a/src/mjlab/entity/data.py
+++ b/src/mjlab/entity/data.py
@@ -122,9 +122,9 @@ class EntityData:
       raise ValueError("Cannot write root COM velocity for fixed-base entity.")
     assert velocity.shape[-1] == self.ROOT_VEL_DIM
 
-    env_ids = self._resolve_env_ids(env_ids)
+    env_ids = env_ids if env_ids is not None else slice(None)
     com_offset_b = self.model.body_ipos[:, self.indexing.root_body_id]
-    quat_w = self.data.qpos[env_ids, self.indexing.free_joint_q_adr[3:7]]
+    quat_w = self.data.qpos[:, self.indexing.free_joint_q_adr[3:7]][env_ids]
     com_offset_w = quat_apply(quat_w, com_offset_b[env_ids])
     lin_vel_com = velocity[:, :3]
     ang_vel_w = velocity[:, 3:]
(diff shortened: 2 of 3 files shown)


kevinzakka:
cc @EGalahad

PR #815, opened 2026-03-18
Generalize RayCastSensor for multi-frame sensing and ring patterns

kevinzakka:
Generalizes `RayCastSensor` to support multiple attachment frames from a single sensor instance, so you can cast rays from e.g. all four feet on a quadruped without creating separate sensors.

**Multi-frame sensing.** `frame` now accepts a tuple of `ObjRef`. The ray pattern is tiled across all frames, each frame's parent body is excluded independently, and output distances are `[B, F*N]`. New `frame_pos_w` and `frame_quat_w` fields on `RayCastData` expose per-frame world poses. `height_scan` automatically uses each frame's Z for terrain-relative height computation.

**Ring pattern.** New `RingPatternCfg` places rays in concentric rings around each attachment frame.

**Direction and origin are now decoupled.** `ray_alignment` controls ray direction only. The ray origin is always the physical frame position (`site_xpos` / `geom_xpos`). Previously (#778), non-base alignment modes also projected the site offset through the alignment rotation to stabilize the origin for trunk-mounted sensors. That projection gives wrong positions for sensors on articulated limbs (e.g. foot sites on bent legs), because it ignores the parent body's rotation. Since every terrain scan config in practice attaches to a body frame (which is unaffected), this change is safe. If a stable projected origin is needed for a site with a large offset, use a body frame instead.

**Per-sensor debug viz toggles.** Raycast sensors with `debug_vis=True` now get individual checkboxes in the Viser viewer's Debug Viz panel, matching the existing pattern for command terms. Toggling while paused triggers an immediate redraw.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -122,6 +122,12 @@ Added
   to bodies with configurable duration and optional application point offset.
 - ONNX auto-export and metadata attachment for manipulation tasks (lift cube)
   on every checkpoint save, matching the velocity and tracking task behavior.
+- Multi-frame ``RayCastSensor``: pass a tuple of ``ObjRef`` to ``frame`` for
+  per-site raycasting with independent body exclusion. New properties:
+  ``num_frames``, ``num_rays_per_frame``. New ``RayCastData`` fields:
+  ``frame_pos_w`` and ``frame_quat_w``.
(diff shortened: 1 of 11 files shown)


mktk1117:
Thanks! LGTM

PR #817, opened 2026-03-19
Add per-foot height sensor for terrain-aware gait rewards

kevinzakka:
Foot height sensing now correctly measures vertical clearance above terrain instead of raw ray distances. This fixes feet_clearance, feet_swing_height, and foot_height observations which were computing incorrect values on non-flat terrain.

Adds RayCastSensor.per_frame_height() to compute true vertical heights (frame_z - hit_z) per frame.

### DIFF
--- a/src/mjlab/sensor/raycast_sensor.py
+++ b/src/mjlab/sensor/raycast_sensor.py
@@ -1,190 +1,10 @@
 """Raycast sensor for terrain and obstacle detection.
 
-Ray Patterns
-------------
+Provides :class:`RayCastSensor` and :class:`RayCastSensorCfg` for BVH-accelerated
+raycasting with grid, pinhole camera, and ring patterns. Supports multi-frame
+attachment, configurable ray alignment, and geom group filtering.
 
-This module provides three ray pattern types for different use cases:
-
-**Grid Pattern** - Parallel rays in a 2D grid::
-
-    Camera at any height:
-          ↓   ↓   ↓   ↓   ↓      ← All rays point same direction
-          ↓   ↓   ↓   ↓   ↓
-          ↓   ↓   ↓   ↓   ↓
-        ──●───●───●───●───●──    ← Fixed spacing (e.g., 10cm apart)
-             Ground
-
-- Rays are parallel (all point in the same direction, e.g., -Z down)
-- Spacing is defined in world units (meters)
-- Height doesn't affect the hit pattern - same footprint regardless of altitude
-- Good for: height maps, terrain scanning with consistent spatial sampling
-
-**Pinhole Camera Pattern** - Diverging rays from a single point::
-
-    Camera LOW:                    Camera HIGH:
-             📷                            📷
-            /|\\                          /  |  \\
-           / | \\                        /   |   \\
-          /  |  \\                      /    |    \\
-        ─●───●───●─                 ───●─────●─────●───
-        (small footprint)           (large footprint)
-
-- Rays diverge from a single point (like light entering a camera)
-- FOV is fixed in angular units (degrees)
-- Higher altitude → wider ground coverage, more spread between hits
-- Lower altitude → tighter ground coverage, denser hits
-- Good for: simulating depth cameras, LiDAR with angular resolution
-
-**Ring Pattern** - Rays in concentric rings around the origin::
-
-    Ring (8 samples):
-         ●   ●
-       ●       ●     ← evenly spaced on circle
-       ●       ●
-         ●   ●
-           ●         ← center (optional)
-
-- Rays cast from offsets arranged in rings around the attachment frame
-- Supports multiple concentric rings at different radii
-- Good for: per-site height sensing, foot clearance, terrain sampling
-
-**Pattern Comparison**:
-
-============== ==================== ========================== =================
-Aspect         Grid                 Pinhole                    Ring
-============== ==================== ========================== =================
-Ray direction  All parallel         Diverge from origin        All parallel
-Spacing        Meters               Degrees (FOV)              Radius + count
-Height affect  No                   Yes                        No
-Real-world     Orthographic proj.   Perspective camera / LiDAR Terrain probe
-============== ==================== ========================== =================
-
-The pinhole behavior matches real depth sensors (RealSense, LiDAR) - when
-you're farther from an object, each pixel covers more area.
-
-
-Frame Attachment
-----------------
-
-Rays are attached to one or more frames in the scene via ``ObjRef``. Supported
-frame types:
-
-- **body**: Attach to a body's origin. Rays follow body position and orientation.
-- **site**: Attach to a site. Useful for precise placement or offset from body.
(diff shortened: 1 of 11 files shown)


src/mjlab/tasks/velocity/mdp/observations.py
```diff
@@ -42,3 +44,22 @@ def foot_contact_forces(env: ManagerBasedRlEnv, sensor_name: str) -> torch.Tenso
...
+    Tensor of shape [B, F] where F is the number of frames (feet).
+  """
+  sensor: RayCastSensor = env.scene[sensor_name]
+  data = sensor.data
+  F, N = sensor.num_frames, sensor.num_rays_per_frame
+  B = data.distances.shape[0]
+  dists = data.distances.view(B, F, N)
+  miss = dists < 0
+  dists = torch.where(miss, torch.full_like(dists, sensor.cfg.max_distance), dists)
+  return dists.min(dim=-1).values  # [B, F]
```
mktk1117: Is this correct? better to double check if there are offsets above. 
Would be better to get the pos of the hit like done in the height scan obs.

src/mjlab/tasks/velocity/mdp/observations.py
```diff
@@ -42,3 +44,22 @@ def foot_contact_forces(env: ManagerBasedRlEnv, sensor_name: str) -> torch.Tenso
   assert sensor_data.force is not None
   forces_flat = sensor_data.force.flatten(start_dim=1)  # [B, N*3]
   return torch.sign(forces_flat) * torch.log1p(torch.abs(forces_flat))
+
+
+def _foot_terrain_height(env: ManagerBasedRlEnv, sensor_name: str) -> torch.Tensor:
```
mktk1117: Having this in obs and using in reward makes it a bit hard to follow. 
Either have this in the reward.py or implement this logic as a sensor... ?
Like have some aggregation method (min, max, average median etc)

src/mjlab/tasks/velocity/mdp/rewards.py
```diff
@@ -193,27 +196,27 @@ class feet_swing_height:
   """Penalize deviation from target swing height, evaluated at landing."""
 
   def __init__(self, cfg: RewardTermCfg, env: ManagerBasedRlEnv):
-    self.sensor_name = cfg.params["sensor_name"]
-    self.site_names = cfg.params["asset_cfg"].site_names
+    self.contact_sensor_name = cfg.params["sensor_name"]
```
louislelay: Isn't this a dead assignment?

src/mjlab/tasks/velocity/mdp/rewards.py
```diff
@@ -167,16 +169,17 @@ def feet_air_time(
...
   asset_cfg: SceneEntityCfg = _DEFAULT_ASSET_CFG,
 ) -> torch.Tensor:
   """Penalize deviation from target clearance height, weighted by foot velocity."""
   asset: Entity = env.scene[asset_cfg.name]
-  foot_z = asset.data.site_pos_w[:, asset_cfg.site_ids, 2]  # [B, N]
+  foot_height = _foot_terrain_height(env, height_sensor_name)  # [B, F]
   foot_vel_xy = asset.data.site_lin_vel_w[:, asset_cfg.site_ids, :2]  # [B, N, 2]
   vel_norm = torch.norm(foot_vel_xy, dim=-1)  # [B, N]
-  delta = torch.abs(foot_z - target_height)  # [B, N]
+  delta = torch.abs(foot_height - target_height)  # [B, N]
```
louislelay: NIT: `foot_height` is annotated [B, F] but delta (derived from it) is annotated [B, N]. Since F and N both mean "number of feet" here, should we pick one notation and use it consistently to avoid confusion?

src/mjlab/tasks/velocity/velocity_env_cfg.py
louislelay, line 1: In the PR description, you mention:

> The spawn Z range is widened from (0.01, 0.05) to (0.05, 0.10) to reduce initial terrain penetration.

But `pose_range` z is still (0.01, 0.05) in the diff. Was this change dropped intentionally or is it still sitting in your stash?

kevinzakka: It's something I had but removed because didn't wanna pollute the PR. Updated the PR description though, thanks Louis.

src/mjlab/tasks/velocity/mdp/rewards.py
kevinzakka: Fixed!

src/mjlab/tasks/velocity/mdp/rewards.py
kevinzakka: Fixed!

src/mjlab/tasks/velocity/mdp/observations.py
kevinzakka: Implemented as a sensor method: `RayCastSensor.per_frame_height()`. Both obs and rewards now call the sensor directly instead of a private helper in obs.

src/mjlab/tasks/velocity/mdp/observations.py
kevinzakka: Done.

kevinzakka:
Thanks @mktk1117 and @louislelay for the review!

src/mjlab/sensor/raycast_sensor.py
mktk1117, line 1: I feel it's too specific to terrain height to put in to the RayCastSensor. 
If we want to use for other directions (for example spherical scan for obstacle avoidance) we want a ray's distance. 
So it might be better to separate this from the RayCastSensor to keep the ray cast sensor to purely get the ray distance, hit pos etc. And more specific feature to other function, or create a custom sensor inheriting raycastsensor.

mktk1117: If we make a thin custom sensor inherited from raycast sensor, we can make, different types of sensors which uses raycasting.

MinDistanceSensor (Take min of rays. useful for obstacle avoidance)
TerrainHeightSensor
DepthCameraSensor (we need to project the ray's distance for depth image conversion)
VoxelSensor
(CeilingHeightSensor...?) etc.

kevinzakka: Fixed.

src/mjlab/sensor/terrain_height_sensor.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+    reduction = self.cfg.reduction
+    if reduction == "min":
+      reduced = heights.min(dim=-1).values
+    elif reduction == "max":
+      reduced = heights.max(dim=-1).values
+    elif reduction == "mean":
+      reduced = heights.mean(dim=-1)
+    else:
```
mktk1117, line 80: sorry, very tiny thing. Maybe add reduction none or raw.

kevinzakka: Done!

PR #835, opened 2026-03-22
Fix TerrainHeightSensor reporting max_distance during ground contact

kevinzakka:
When a foot penetrates slightly below the terrain surface, all downward rays miss and the sensor reported max_distance instead of ~0. This corrupted foot clearance rewards and caused heel-walking gaits.

The fix uses frame_z clamped to [0, max_distance] as the fallback for all-miss frames instead of unconditionally returning max_distance.


hsb0508:
截屏2026-03-23 下午3 31 53
Still has some issue with the height sensor in rought terrain. When the feet is slightly below the terrain the sensor will report the height from lower surface of the box.

kevinzakka:
@hsb0508 I'll take a look, thanks for reporting.

PR #841, opened 2026-03-24
Fix TerrainHeightSensor reporting box thickness during penetration

### LINKED ISSUE #838: TerrainHeightSensor reports height from bottom face of box terrain during penetration
#835 fixed the case where all rays miss during foot penetration, but there's a second failure mode on box terrain: when a foot is slightly below the surface, downward rays originate inside the box and hit its bottom face. This produces a large bogus height (the box thickness) instead of ~0.

The root cause is the same as #835 (ray origin below terrain surface), but manifests differently with box geoms vs planes. Planes are one-sided so rays miss entirely; boxes have a physical bottom face that returns a valid hit. Heightfields and meshes use two-sided triangles, so a ray from slightly below hits the same triangle at near-zero distance, making the error negligible. The large bogus height is box-specific.

Reported in https://github.com/mujocolab/mjlab/pull/835#issuecomment-4114146045.

kevinzakka:
When a foot penetrates box terrain, downward rays originate inside the box and hit its bottom face. The sensor reports the box thickness as the height instead of ~0. This is different from the plane penetration case fixed in #835, where rays miss entirely. With boxes, the rays do hit something, just the wrong surface.

The fix checks the hit normal: if a ray hit a downward-facing surface (normal_z < 0), the ray origin is inside geometry and clearance is zero. This is distinct from true misses (distance < 0), which continue to use the existing frame_z fallback.

Also consolidated TerrainHeightSensor tests into a single file and removed redundant/vacuous tests from the previous PR.

Fixes #838


kevinzakka:
cc @@hsb0508

PR #862, opened 2026-03-30
Add per-pixel segmentation camera output and multi-cube manipulation task

pthangeda:
Adds segmentation (geom ID per pixel) as a camera data type alongside RGB and depth, and a multi-cube goal-conditioned lifting task that uses it. The mujoco_warp side already landed upstream (google-deepmind/mujoco_warp#1236). 

---------

**Sensor layer:** `"segmentation"` is a new `CameraDataType`. `SensorContext` wraps the warp seg buffer as a zero-copy torch view, same pattern as depth. Shape is `[B, H, W, 1]` int32 — geom IDs for hits, -1 for background, -2 for flex. Zero cost when not requested.

**Observation terms:** Three new functions — `camera_segmentation` (raw IDs), `camera_entity_mask` (binary mask for any named entity via geom ID lookup), and `camera_target_cube_mask` (binary mask for whichever cube the command selected as the current target).

**Multi-cube task:** `MultiCubeLiftingCommand` randomly picks 1 of N cubes as the target each episode, randomizes all cube poses, and exposes `target_geom_ids` for the segmentation mask. Registered as `Mjlab-Multi-Cube-Seg-Yam` with 3 cubes. The policy receives depth + target binary mask as a 2-channel image through a spatial softmax CNN.

**Viser viewer:** Colorized segmentation feed in the Camera Feeds pane, following the same pattern as RGB and depth. One thing I'm not fully happy with — the viewer shows all geom IDs colored, not the actual policy input (the binary target mask). Since camera viewer operates at the sensor level and afaik no existing mechanism to inject task-level observation visualizations into it, I think the right fix is a separate observation visualization system and I'll follow up with a PR for that.

**Tests:** Shape, content, and combined (RGB+depth+seg) tests matching the existing camera test patterns. Successfully trained policy on the sample multi-cube task - I can try and add non-manipulation envs that use segmentation obs too if needed. 

https://github.com/user-attachments/assets/e96a2931-5b5d-4673-b1a0-ed8f048d84b6

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -39,6 +39,10 @@ Added
 - Added a Checkpoints tab to the Viser play viewer for hot-swapping
   checkpoints without restarting. Works with local directories and W&B
   runs (:issue:`751`). Contribution by @omarrayyann.
+- Added ``"segmentation"`` camera data type for per-pixel geom ID output
+  alongside RGB and depth, and a multi-cube goal-conditioned lifting task
+  (``Mjlab-Multi-Cube-Seg-Yam``) that uses it (:issue:`862`).
+  Contribution by @pthangeda.
 
 Changed
 ^^^^^^^


kevinzakka:
@pthangeda Can I push changes to your branch?

e.g., gonna delete `camera_entity_mask` since it's unused and clean up a few other nits.

pthangeda:
@kevinzakka yes go for it! (also feel free to do it in any of my future PRs without waiting - I always keep edit access to maintainers enabled)

pthangeda:
https://github.com/user-attachments/assets/eeb5e992-f102-4c8e-861c-7840b675a7dc

Another video with a few more cubes (:

PR #872, opened 2026-04-01
Observation and action extractor

ManuelActisCassa:
Add the option to extract observation and action per frame during inference.
The data is formatted and written into a csv file.
csv  files are saved into {mjlab parent directory} / mjlab / rollout_obs_csv /

The option is only supported when running a single environment (num-envs = 1).

This data could be used to observe system evolution or to train ML tools (state predictors, state estimators...)

**To use it** :
uv run {mjlab task} --csv-save-file {str} --csv-chunk-save-size {int} --{other options}

**Options** :
csv-save-file (str) : name of the file that will be created to save the collected data (don't pass ".csv")
csv-chunk-save-size (int) : the data is saved in the csv file in chunks, to protect the computer disk. Minimum is 200, default is 500.

**Example** : 

uv run play Mjlab-Velocity-Flat-Unitree-G1 --csv-save-file extracted_dataset --csv-chunk-save-size 300 --agent random

Results are saved in : {mjlab parent directory} / mjlab / rollout_obs_csv / extracted_dataset.csv

image

### DIFF
--- a/src/mjlab/envs/manager_based_rl_env.py
+++ b/src/mjlab/envs/manager_based_rl_env.py
@@ -1,9 +1,7 @@
+import csv
 import math
 from dataclasses import dataclass, field
 from typing import Any
-
-import csv
-from pathlib import Path
 
 import mujoco
 import numpy as np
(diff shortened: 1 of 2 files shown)


kevinzakka:
Hi @ManuelActisCassa, #884 should solve this more generally.

PR #902, opened 2026-04-12
Add auto_reset config flag

### LINKED ISSUE #900: Making auto-reset optional for mjlab
Hi Kevin / mjlab team,

Thanks for the great work on mjlab — really enjoying the framework so far.

I noticed that `env.step()` currently **automatically resets environments** when they are terminated or truncated, and returns the **observation after reset**. This makes it hard to access the true next state $s_{t+1}$ at the end of an episode.

This is problematic for some RL algorithms. For example, when computing targets like:

$$
Q(s_t, a_t) = r_t + \gamma Q(s_{t+1}, a_{t+1})
$$

we still need the real $s_{t+1}$, especially in truncation cases. With auto-reset, that state is not available.

This also differs from frameworks like TorchRL:

TorchRL explicitly distinguishes auto-reset vs non-auto-reset environments. In auto-reset environments, the returned observation may actually belong to the next episode, not the terminal transition. 

More importantly, the TorchRL maintainers have raised concerns about the gymnasium 1.0 interface, which contains similar auto-reset behaviors [here](https://github.com/pytorch/rl/discussions/2483)

### **Suggestion**

It would be helpful to add an option to disable auto-reset, e.g.:

```python
env = make_env(..., auto_reset=False)
# or, alternatively
env = ManagerBasedRlEnv(..., auto_reset=False)
```

When disabled:

* `step()` should return the true terminal/truncated observation
* Reset should be handled explicitly via `env.reset(...)`

Keeping the current behavior as default would preserve backward compatibility.

---

I’d be happy to help implement this as a PR if this sounds reasonable.

Thanks!

kevinzakka:
Adds `ManagerBasedRlEnvCfg.auto_reset` (default `True`). When `False`, `step()` skips the reset block for done environments and returns the true terminal observation; the caller must call `reset(env_ids=...)` before the next `step()` or the env raises. Algorithms that need the real terminal state (off-policy Q targets, correct PPO truncation bootstrap) get direct access to it.

Note that mjlab's default `train.py` path goes through rsl_rl's `OnPolicyRunner`, which doesn't drive manual resets, so `auto_reset=False` is intended for users running their own training loop or a wrapper that handles the reset between steps.

An earlier version of this branch also set `terminal_observation` in `extras` under `auto_reset=True`. Getting that right would have required broader changes across the manager layer, and I wanted to keep this PR self-contained. Users who need terminal-obs bootstrap can opt into `auto_reset=False` and own the reset in a thin wrapper.

Closes #900.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -8,6 +8,13 @@ Upcoming version (not yet released)
 Added
 ^^^^^
 
+- Added ``ManagerBasedRlEnvCfg.auto_reset`` flag. When ``True`` (default),
+  ``step()`` continues to reset done environments in place and returns the
+  post-reset observation. When ``False``, ``step()`` skips the reset block
+  and returns the terminal observation directly; the caller must call
+  ``reset(env_ids=...)`` for done environments before the next ``step()``
+  or a ``RuntimeError`` is raised. Enables access to the true terminal
+  state for algorithms that need it (:issue:`900`).
 - Added ``ActuatorCfg.viscous_damping`` for passive velocity proportional
   damping (``f = -b·v``), distinct from the PD derivative gain ``damping``
   used by position and velocity actuators. Maps to ``<joint damping>`` for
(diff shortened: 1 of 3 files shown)


saikishor (commented):
LGTM. 

Maybe it is worth mentioning in the code docs or changelog about the mjlab going through rsl_rl's OnPolicyRunner?

kevinzakka:
@saikishor Done!

saikishor (approved):
Spot on!

PR #937, opened 2026-04-24
Per-world mesh variants: explicit inertials + variant-aware DR and viewers

XiangruiJiang:
Summary                                                         
         
  Builds on per-world-mesh to (1) let mesh variants carry their own explicit body inertials through per-world compilation, and (2) fix two bugs that per-world variants exposed: domain randomization reading defaults with the wrong index convention, and both the
  native/offscreen/viser viewers rendering every world with variant-0 geometry.                                                                                                                                                                                                    
  
  1. Per-variant explicit inertials — entity/entity.py, scene/per_world_mesh.py                                                                                                                                                                                                    
                                                                  
  - New frozen dataclass BodyInertialMetadata(body_name, mass, ipos, inertia, iquat).                                                                                                                                                                                              
  - VariantMetadata gains variant_body_inertials: tuple[tuple[BodyInertialMetadata, ...], ...].
  - During _build_merged_spec, _collect_explicit_body_inertials walks each variant body tree (_iter_body_tree) and snapshots every body with explicitinertial=1. Names are stored local to the variant spec; per_world_mesh prefixes them with the scene entity name when applying.
  - _populate_dependent_fields now takes world_to_variant. For every unique world-variant key it does restore → apply variant inertials → compile, then restores once more at the end. Each variant body's explicitinertial, mass, ipos, inertia, iquat, fullinertia are saved     
  before the loop and reset between compiles so variants don't bleed into each other.                                                                                                                                                                                              
  - The previous if len(unique_rows) <= 1: return shortcut is removed — with explicit inertials, even a single unique variant still needs the apply/compile pass or the saved inertials never hit the spec.                                                                        
  - Validated on 024_bowl: compiled mass ≈ 0.01645 kg (matches the provided inertial metadata) vs. MuJoCo's mesh-density estimate of ≈ 0.554 kg; also verified on multi-env scenes where each env reports its own variant mass.                                                    
                                                                                                                                                                                                                                                                                   
  2. DR default-value indexing fix — envs/mdp/dr/_core.py, envs/mdp/dr/body.py                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                   
  Previously:                                                                                                                                                                                                                                                                      
  default_field[entity_indices].unsqueeze(0).expand_as(indexed_data)
  On standard defaults (shape (nentity, …)) this was fine. On per-world defaults (shape (nworld, nentity, …), produced by per-world mesh compilation) it silently indexed the world dimension with entity indices, then broadcast that wrong slice across envs.
                                                                                                                                                                                                                                                                                   
  Fix: new helper                                                                                                                                                                                                                                                                  
  _select_default_values(env, field, env_ids, entity_indices) -> (n_env, n_entity, …)                                                                                                                                                                                              
  uses torch.meshgrid(env_ids, entity_indices, indexing="ij") when the field is in sim.per_world_default_fields, otherwise indexes entities and broadcasts envs. _randomize_model_field, _randomize_quat_field, and pseudo_inertia all route through it; the now-redundant         
  L.unsqueeze(0).expand(...) inside pseudo_inertia is dropped since L already carries the env dim.                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                   
  3. Shared viewer model-sync module — new viewer/_model_sync.py                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                   
  Extracts from native/viewer.py:                                 
  - disable_model_sameframe_shortcuts(model) — clears body_sameframe, geom_sameframe, site_sameframe, and body_simple so host-side mj_forward stops taking the compile-time shortcut that mujoco_warp's kernels don't emulate. Necessary whenever per-world variants change local  
  offsets.                                                                                                                                                                                                                                                                         
  - sync_visual_fields(dst_model, sim_model, expanded_fields, env_idx) — copies expanded_fields ∩ VISUAL_FIELDS from GPU into a host MjModel.
  - Module-level INERTIAL_FIELDS and VISUAL_FIELDS frozensets.                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                   
  VISUAL_FIELDS gains body_subtreemass on top of the set in native/viewer.py. mj_forward uses it as the denominator when normalizing subtree_com; without syncing it, mjVIS_COM decor and anything reading data.subtree_com lands at wrong positions as soon as body_mass varies   
  per-world.                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                   
  native/viewer.py drops its duplicates and imports from the new module.                                                                                                                                                                                                           
                                                                  
  4. Offscreen renderer + Viser become variant-aware                                                                                                                                                                                                                               
                                                                  
  viewer/offscreen_renderer.py — OffscreenRenderer.__init__ accepts optional sim_model and expanded_fields; when provided it disables sameframe shortcuts up front and update() calls _sync_model_fields(env_idx) before the main env, again before each neighbor env rendered for 
  context, and once more at the end to restore the main env's view. manager_based_rl_env.py wires sim.model / sim.expanded_fields into the constructor.
                                                                                                                                                                                                                                                                                   
  viewer/viser/scene.py — MjlabViserScene.__init__ takes the same two args. When expanded_fields overlap with _GEOMETRY_HANDLE_FIELDS (geom_dataid, geom_rgba, geom_size, geom_pos, geom_quat, mat_rgba), it switches to a per-variant handle path:                                
  
  - New _create_mesh_handles_by_group override iterates every env, syncs visual fields, and groups dynamic-body geoms by _geom_subgroup_visual_fingerprint (type, dataid, matid+rgba, size, rgba, pos, quat). Each fingerprint becomes one _PerWorldMeshGroup batched handle       
  covering only the envs that share it — no more variant-0 geometry shown for every env.
  - New _update_visualization_locked override drives those per-world handles with (env_ids, body_ids) gathers on body_xpos/body_xmat (and mocap positions where applicable), with a show_only_selected mask path for single-env focus mode.                                        
  - Non-variant geoms still use the base class path.                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                   
  viewer/viser/viewer.py wires sim.model / sim.expanded_fields into the scene.

### DIFF
--- a/src/mjlab/entity/entity.py
+++ b/src/mjlab/entity/entity.py
@@ -331,8 +331,8 @@ class Entity:
         geom = template_body.add_geom()
         geom.type = mujoco.mjtGeom.mjGEOM_MESH
         geom.meshname = f"{longest_prefix}{longest_names[k]}"
-        geom.contype = 0
-        geom.conaffinity = 0
+        geom.contype = 1
+        geom.conaffinity = 1
 
     # Build variant_mesh_names: use original names with variant prefix.
     variant_mesh_name_lists: list[tuple[str | None, ...]] = []
--- a/src/mjlab/entity/entity.py
+++ b/src/mjlab/entity/entity.py
@@ -77,6 +77,17 @@ class VariantCfg:
   weight: float = 1.0
 
 
+@dataclass(frozen=True)
+class BodyInertialMetadata:
+  """Explicit inertial properties for one body in a mesh variant."""
+
+  body_name: str
+  mass: float
+  ipos: tuple[float, float, float]
+  inertia: tuple[float, float, float]
+  iquat: tuple[float, float, float, float]
+
+
 @dataclass
 class VariantMetadata:
   """Bookkeeping produced by Entity when merging variant specs."""
@@ -87,6 +98,34 @@ class VariantMetadata:
   # have None for padding slots that should be disabled (dataid = -1).
   variant_mesh_names: tuple[tuple[str | None, ...], ...]
   num_mesh_geoms: int  # Max mesh geom count after padding.
+  # Per-variant explicit body inertials. Names are local to the variant spec;
+  # per_world_mesh prefixes them with the scene entity name when applying them.
+  variant_body_inertials: tuple[tuple[BodyInertialMetadata, ...], ...] = ()
+
+
+def _iter_body_tree(body: mujoco.MjsBody):
+  yield body
+  for child in body.bodies:
+    yield from _iter_body_tree(child)
+
+
+def _collect_explicit_body_inertials(
+  root_body: mujoco.MjsBody,
+) -> tuple[BodyInertialMetadata, ...]:
+  inertials: list[BodyInertialMetadata] = []
+  for body in _iter_body_tree(root_body):
+    if not body.name or not body.explicitinertial:
+      continue
+    inertials.append(
+      BodyInertialMetadata(
+        body_name=body.name,
+        mass=float(body.mass),
+        ipos=tuple(float(x) for x in body.ipos),
+        inertia=tuple(float(x) for x in body.inertia),
+        iquat=tuple(float(x) for x in body.iquat),
+      )
+    )
+  return tuple(inertials)
 
 
 @dataclass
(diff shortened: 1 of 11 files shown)


kevinzakka (approved):
Thanks @XiangruiJiang! Going to do a followup rebase + clean up + simplification pass but this is a great PR thank you.

XiangruiJiang:
Thank you very much, and I'm sorry my PR wasn't good enough

PR #939, opened 2026-04-27
Add --log-root CLI option to train, play, and evaluate

louislelay:
The log root was hardcoded, which forces workarounds in repos that have a designated output directory. This adds a --log-root option to train, play, and evaluate so the user can point logs to the right place. The default is unchanged (logs/rsl_rl), so existing workflows keep working.


kevinzakka (approved):
Thanks @louislelay!

PR #948, opened 2026-04-28
Raycast sensor optimization

bd-pdomanico:
Removes boolean-mask indexing operations with `masked_fill_` and an implementation of hit_pos_w that uses a clamped distance to put hit_pos_w for misses at world_origins. These effectively remove all cuda syncs from the ray postprocess, unblocking the cpu thread while gpu-based sensing occurs.

Also optimizes `quat_from_matrix`. Similarly this removes cuda sync operations and catches the implementation up to the latest from https://github.com/facebookresearch/pytorch3d/blob/main/pytorch3d/transforms/rotation_conversions.py. Profiling now shows:

    device=NVIDIA GeForce RTX 4090

    shape              legacy us   current us   speedup     max |Δ|
    ---------------------------------------------------------------
    B = 4096              231.26        51.01     4.53x    1.79e-07
    B = 16384             243.55        56.07     4.34x    1.79e-07
    B = 65536             282.05        53.93     5.23x    1.79e-07
    
Profiled using: 
```python
def time_call(fn, x, *, warmup=20, iters=200) -> float:
    """Mean per-call latency in microseconds, measured with cudaEvents."""
    for _ in range(warmup):
        fn(x)
    torch.cuda.synchronize()
    starts = [torch.cuda.Event(enable_timing=True) for _ in range(iters)]
    ends = [torch.cuda.Event(enable_timing=True) for _ in range(iters)]
    for i in range(iters):
        starts[i].record()
        fn(x)
        ends[i].record()
    torch.cuda.synchronize()
    times_ms = [s.elapsed_time(e) for s, e in zip(starts, ends)]
    return statistics.mean(times_ms) * 1000.0  # us

m = random_rotations(shape, device=device, dtype=dtype)

# Numerical agreement (account for ±q sign ambiguity)
with torch.no_grad():
    ql = quat_from_matrix_legacy(m)
    qc = quat_from_matrix_current(m)
    sign = torch.sign((ql * qc).sum(dim=-1, keepdim=True))
    sign = torch.where(sign == 0, torch.ones_like(sign), sign)
    max_diff = (ql - qc * sign).abs().max().item()

t_legacy = time_call(lambda x: quat_from_matrix_legacy(x), m)
t_current = time_call(lambda x: quat_from_matrix_current(x), m)
speedup = t_legacy / t_current
```

With both of these changes, `Mjlab-Velocity-Rough-Unitree-Go1` trains ~3% faster

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -25,6 +25,8 @@ Added
 Changed
 ^^^^^^^
 
+- Sped up ``RayCaster`` post-processing and ``quat_from_matrix``. 
+  ``quat_from_matrix`` output may differ by ~2e-7 vs. the previous implementation.
 - Bumped ``rsl-rl-lib`` from 5.0.1 to 5.2.0. This brings ``torch.compile`` support for
   PPO and Distillation, and optional std clamping and constant std in
   ``GaussianDistribution``. No code changes required on the mjlab side.
--- a/src/mjlab/sensor/raycast_sensor.py
+++ b/src/mjlab/sensor/raycast_sensor.py
@@ -781,17 +781,18 @@ class RayCastSensor(Sensor[RayCastData]):
     assert self._ray_dist is not None and self._ray_normal is not None
     distances = wp.to_torch(self._ray_dist)
     normals_w = wp.to_torch(self._ray_normal).view(B, self._num_rays, 3)
-    distances[distances > self.cfg.max_distance] = -1.0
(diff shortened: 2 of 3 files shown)


kevinzakka (changes requested):
This is awesome, thanks @bd-pdomanico! Two small asks before merging:

1. Rebase onto the latest main
2. Add a changelog entry

Thanks!

kevinzakka (approved):
Thank you!

PR #950, opened 2026-04-28
Skip Warp CUDA init in seed_rng when device is CPU

### LINKED ISSUE #949: `ManagerBasedRlEnv` unconditionally initializes warp on all visible CUDA devices
## Issue

I’d like to be able to initialize a `ManagerBasedRlEnv` in a way that it doesn’t use the GPU at all, even if a GPU is visible. I would expect that passing `device=”cpu”` would accomplish this, but it doesn't: `device=”cpu”` puts all computation on the CPU, but warp still initializes on the GPU.

I traced the issue to the `wp.rand_init(wp.in32(seed))` call in `mjlab.utils.random.seed_rng()`. This is called by `ManagerBasedRlEnv.__init__()` if `cfg.seed` is provided, regardless of device. It appears to initialize warp on all visible devices.

## Steps to reproduce

I wrote this test that checks for Warp CUDA initialization when a `ManagerBasedRlEnv` is initialized with `device="cpu"`:
```
import warp as wp
from mjlab.envs import ManagerBasedRlEnv
from mjlab.tasks.cartpole.cartpole_env_cfg import cartpole_balance_env_cfg


def test_manager_based_rl_env_with_seed_does_not_initialize_cuda_on_cpu_device() -> (None):
    assert wp._src.context.runtime is None, ("Warp must not be initialized before this test")
    cfg = cartpole_balance_env_cfg()
    cfg.scene.num_envs = 1
    cfg.seed = 42
    env = ManagerBasedRlEnv(cfg=cfg, device="cpu")
    env.close()
    cuda_devices = [d for d in wp.get_devices() if "cuda" in str(d)]
    assert not cuda_devices, (
        f"ManagerBasedRlEnv(device='cpu', seed=42) initialized Warp CUDA devices "
        f"{cuda_devices}")
```
When I run this on a machine with GPU visible, I get the following failure. This shows that `device="cpu"` does not fully prevent GPU usage.
```
AssertionError: ManagerBasedRlEnv(device='cpu', seed=42) initialized Warp CUDA devices ['cuda:0']
```
----
This test shows that `seed_rng` initializes Warp on all visible devices. It also fails for me, indicating that `seed_rng` is the issue.
```
import warp as wp
from mjlab.utils.random import seed_rng


def test_seed_rng_does_not_initialize_cuda_on_cpu_device() -> None:
    assert wp._src.context.runtime is None, ("Warp must not be initialized before this test")
    seed_rng(42)
    rt = wp._src.context.runtime
    cuda_devices = [d for d in wp.get_devices() if "cuda" in str(d)]
    assert rt is None or not cuda_devices, (
        f"seed_rng() initialized Warp CUDA devices {cuda_devices} ")
```

kevinzakka:
ManagerBasedRlEnv unconditionally called wp.rand_init via seed_rng, which initializes Warp on every visible CUDA device. So constructing the env with device="cpu" still claimed a CUDA context on machines with a visible GPU.

seed_rng now takes an optional device argument and skips wp.rand_init when the device resolves to CPU. ManagerBasedRlEnv passes its constructor device through.

Fixes #949

### DIFF
--- a/tests/test_random.py
+++ b/tests/test_random.py
@@ -1,0 +1,30 @@
+"""Tests for mjlab.utils.random."""
+
+import subprocess
+import sys
+import textwrap
+
+
+def test_seed_rng_cpu_device_does_not_initialize_warp_cuda() -> None:
+  """seed_rng(device="cpu") must not initialize Warp's CUDA runtime.
+
+  Runs in a subprocess so that Warp is guaranteed uninitialized before the
+  call.
+  """
+  script = textwrap.dedent("""
+    import warp as wp
+    from mjlab.utils.random import seed_rng
+
+    assert wp._src.context.runtime is None, "Warp must not be initialized yet"
+    seed_rng(42, device="cpu")
+    rt = wp._src.context.runtime
+    if rt is not None:
+        cuda = [d for d in wp.get_devices() if "cuda" in str(d)]
+        assert not cuda, f"seed_rng(device='cpu') initialized CUDA devices {cuda}"
+  """)
+  result = subprocess.run(
+    [sys.executable, "-c", script], capture_output=True, text=True
+  )
+  assert result.returncode == 0, (
+    f"subprocess failed:\nstdout={result.stdout}\nstderr={result.stderr}"
+  )
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -48,6 +48,11 @@ Changed
 Fixed
 ^^^^^
 
+- Fixed ``ManagerBasedRlEnv`` initializing Warp on all visible CUDA devices
+  even when constructed with ``device="cpu"``. ``seed_rng`` now accepts a
+  ``device`` argument and skips ``wp.rand_init`` on CPU devices, so a
+  CPU-only env no longer claims a CUDA context on machines with a visible
+  GPU (:issue:`949`).
 - Fixed ``ContactSensor.compute_first_contact`` and ``compute_first_air``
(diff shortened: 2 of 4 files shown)


kevinzakka:
The nightly version of mujoco (v3.7.0.dev892927987) was somehow yanked and CI is failing because it does not exist anymore...

tests/test_random.py
```diff
@@ -0,0 +1,30 @@
...
+  script = textwrap.dedent("""
+    import warp as wp
+    from mjlab.utils.random import seed_rng
+
+    assert wp._src.context.runtime is None, "Warp must not be initialized yet"
+    seed_rng(42, device="cpu")
+    rt = wp._src.context.runtime
+    if rt is not None:
+        cuda = [d for d in wp.get_devices() if "cuda" in str(d)]
+        assert not cuda, f"seed_rng(device='cpu') initialized CUDA devices {cuda}"
```
bd-hberger, line 23: I didn't catch this at the time, but this isn't a useful check. `wp.get_devices` returns all visible devices, so this is actually checking if the GPU is visible, not if CUDA has been initialized on it. The relevant question is just whether CUDA has been initialized at all, and `if rt is not None` is sufficient to check that.

bd-hberger (commented):
After more investigation of CUDA initialization, I'm no longer sure that these changes are valuable. `wp.get_devices()` is always going to initialize CUDA on all visible devices, regardless of what `seed_rng` does, and I don't know of a good way around it other than setting the `CUDA_VISIBLE_DEVICES` environment variable

PR #954, opened 2026-04-28
Revert quat_from_matrix decorator from torch.compile to torch.jit.script

kevinzakka:
The @torch.compile(fullgraph=True) decorator added to quat_from_matrix in #948 hits dynamo's recompile_limit (8) when the full test suite runs, causing test_raycast_sensor.py to fail across the board with FailOnRecompileLimitHit. Shape specialization on the leading batch dim exhausts the budget across the many env constructions that happen earlier in the suite, so by the time raycast tests run the cache is poisoned and every test errors out.

Switching back to @torch.jit.script sidesteps dynamo entirely and gets the suite green again. The pytorch3d-style sync-free rewrite from #948 — which is the actual source of the speedup — is preserved untouched; only the decorator changes. The matching changelog entry is dropped since the perf attribution is now muddier.

We can investigate a proper jit → compile switch later (probably with dynamic=True so the batch dim doesn't trigger recompiles, or a per-process dynamo reset in the test harness).

cc @bd-pdomanico

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -37,10 +37,6 @@ Changed
   formulation of ``hit_pos_w`` that places misses at the world origin. This
   removes all CUDA syncs from the ray post-process, letting the CPU thread
   proceed while GPU-based sensing runs. Contribution by @bd-pdomanico.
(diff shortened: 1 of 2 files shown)


bd-pdomanico:
Thanks for fixing, @kevinzakka

PR #960, opened 2026-04-29
Fix extras["log"] entries from reward terms being discarded on reset

### LINKED ISSUE #957: Values written to extras["log"] discarded before return
Currently, the MjLab mdp (reward function) building blocks include several instances of patterns like this
https://github.com/mujocolab/mjlab/blob/0040979763ab43bc1220812c9de4bc74e2631f42/src/mjlab/tasks/velocity/mdp/rewards.py#L352
where arbitrary manager terms inject metrics into the extras log dictionary.

The sequence of operations in the step function is
https://github.com/mujocolab/mjlab/blob/0040979763ab43bc1220812c9de4bc74e2631f42/src/mjlab/envs/manager_based_rl_env.py#L446-L453

where the call to `_reset_idx()` does
https://github.com/mujocolab/mjlab/blob/0040979763ab43bc1220812c9de4bc74e2631f42/src/mjlab/envs/manager_based_rl_env.py#L555-L567

Note the final line `self.extras["log"] = dict()`.

As a result, for any step where a nonzero number of environments terminate, all metrics written using the `env.extras["log"][...] = ...` pattern from reward terms are deleted

## Discussion

My personal opinion is that we should support logging metrics from reward terms like the example velocity reward functions do. It more flexible and allows for efficient reuse of intermediate computation values in a way that is not possible in the metrics manager (which also limits you to a single metric per term). I would recommend moving the clearing of the extras dictionary to the top of step() and reset() but acknowledge that that would diverge the mjlab env behavior from that of Isaaclab.

kevinzakka:
Reward terms (e.g. velocity task metrics like `Metrics/air_time_mean`, `Metrics/slip_velocity_mean`) write into `extras["log"]` during `reward_manager.compute()`, but `_reset_idx()` was clearing the dict with `self.extras["log"] = dict()` before repopulating it with manager reset data. Since `_reset_idx` runs after `reward_manager.compute()` in `step()`, all reward-term metrics were silently discarded on any step where at least one environment resets — which is nearly every step in practice. This meant rsl_rl's logger never saw these metrics in W&B/Neptune/TensorBoard.

The fix moves the `extras["log"]` clearing from `_reset_idx()` to the top of `step()` and `reset()`, so reward terms write after the clear and `_reset_idx` merges additively via `.update()`. No key collisions are possible since manager resets use `Episode_Reward/`, `Episode_Termination/`, etc. while reward terms use `Metrics/`.

Fixes #957

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -75,6 +75,12 @@ Changed
 Fixed
 ^^^^^
 
+- Fixed ``extras["log"]`` entries written by reward terms (e.g. ``Metrics/*``
+  values in velocity tasks) being silently discarded on any step where at
+  least one environment resets. ``_reset_idx`` was clearing the dict after
+  ``reward_manager.compute()`` had already populated it. The clear now
+  happens at the top of ``step()`` and ``reset()`` so that all entries
+  survive (:issue:`957`).
 - Fixed ``ManagerBasedRlEnv`` initializing Warp on all visible CUDA devices
   even when constructed with ``device="cpu"``. ``seed_rng`` now accepts a
   ``device`` argument and skips ``wp.rand_init`` on CPU devices, so a
--- a/src/mjlab/envs/manager_based_rl_env.py
+++ b/src/mjlab/envs/manager_based_rl_env.py
@@ -365,6 +365,7 @@ class ManagerBasedRlEnv:
       env_ids = torch.arange(self.num_envs, dtype=torch.int64, device=self.device)
     if seed is not None:
       self.seed(seed)
+    self.extras["log"] = dict()
     self._reset_idx(env_ids)
     self.scene.write_data_to_sim()
     self.sim.forward()
@@ -414,6 +415,7 @@ class ManagerBasedRlEnv:
         "reset(env_ids=...) before calling step() again when auto_reset=False."
       )
 
+    self.extras["log"] = dict()
     self.action_manager.process_action(action.to(self.device))
 
     for _ in range(self.cfg.decimation):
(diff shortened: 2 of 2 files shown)


kevinzakka:
cc @bd-pdomanico

bd-pdomanico:
Thanks for another quick fix, @kevinzakka!

qiayuanl:
Heads-up for users running mjlab with **rsl_rl**: this clearing-timing change interacts with a latent issue in rsl_rl's `Logger` ([`rsl_rl/utils/logger.py:161`](https://github.com/leggedrobotics/rsl_rl/blob/master/rsl_rl/utils/logger.py) in 5.2.0):

```python
for key in self.ep_extras[0]:
```

It samples keys only from the **first** `ep_info` of the logging window. Pre-#960, `extras["log"]` carried the previous reset's data forward across all steps, so `ep_extras[0]` always had keys. Post-#960, the dict is empty unless an env resets during that exact step. On tasks where no reward term writes to `extras["log"]` per step (e.g. tracking — velocity is fine, it writes every step), iterations whose first step has no reset silently drop all subsequent `Episode_Reward/*`, `Metrics/*`, `Episode_Termination/*` from both the console log and W&B.

Just leaving a breadcrumb for anyone hitting intermittent missing keys after rebasing onto this — not asking for a change here.

bd-pdomanico:
I'm okay reverting since the episode rewards/metrics are established features and the per-step metrics are a "new" pattern in the velocity mjlab examples. 

If it were all up to me, I would try to transition to favor per-step log values and move away from the reset-only, end-of-episode metrics from IsaacLab. For infinite horizon tasks, averaging over end-of-episode metrics and rewards biases the values towards the shorter episodes (e.g. unsuccessful). Unfortunately, that sort of change would be disruptive to anyone that has already established intuition on the current values and tuning of their environment.

PR #963, opened 2026-04-30
[Bug Fix] Fix contact sensor global frame rotation

bd-pdomanico:
There exists a bug when `ContactSensor` `global_frame=True` and `reduce != "netforce"`

Contact forces were incorrectly ordered as [tangent, tangent, normal]. From the mujoco [docs](https://mujoco.readthedocs.io/en/stable/computation/index.html#contact):

> The first ( axis of this frame is the contact normal direction, while the remaining y and z axes define the tangent plane. One might have expected the normal to correspond to the z axis, as in MuJoCo’s visualization convention, but we support frictionless contacts where only the normal axis is used, which is why we want to have the normal in first position.

Adds a unit test that prior to fix would fail with the following but is now passing.
```
>     assert torch.all(sensor_force[:, 1].abs() < 0.05), (
        f"sensor_force y-component should be ~0, got {sensor_force[:, 1].tolist()}"
      )
E     AssertionError: sensor_force y-component should be ~0, got [-2.4564199447631836, -2.4564199447631836]
E     assert tensor(False, device='cuda:0')
E      +  where tensor(False, device='cuda:0') = <built-in method all of type object at 0x7240481dcac0>(tensor([2.4564, 2.4564], device='cuda:0') < 0.05)
E      +    where <built-in method all of type object at 0x7240481dcac0> = torch.all
E      +    and   tensor([2.4564, 2.4564], device='cuda:0') = <built-in method abs of Tensor object at 0x723e8d3ccf50>()
E      +      where <built-in method abs of Tensor object at 0x723e8d3ccf50> = tensor([-2.4564, -2.4564], device='cuda:0').abs
```

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -80,7 +80,7 @@ Fixed
   rotated onto the wrong axis. The contact-frame→world rotation matrix had
   its columns ordered ``[tangent, tangent2, normal]`` instead of
   ``[normal, tangent, tangent2]``, projecting the normal-force component
-  onto a tangent direction.
+  onto a tangent direction. Contribution by @bd-pdomanico.
 - Fixed ``extras["log"]`` entries written by reward terms (e.g. ``Metrics/*``
   values in velocity tasks) being silently discarded on any step where at
   least one environment resets. ``_reset_idx`` was clearing the dict after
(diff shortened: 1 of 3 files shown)


kevinzakka (approved):
Thank you again @bd-pdomanico!

PR #964, opened 2026-05-01
Add Unitree H1 to available robots

paLeziart:
Considering how widespread the Unitree H1 is in the literature or in benchmarks (like HumanoidBench), I felt it would be relevant for the community to have an easy access to this robot, whether for comparison with other approaches or for reproducibility.

This pull request integrates the Unitree H1 in Mjlab.

- I retrieved meshes and the robot .xml from the [Mujoco menagerie](https://github.com/google-deepmind/mujoco_menagerie/tree/main/unitree_h1) 
- I adapted the .xml taking inspiration from G1's one to fit mjlab, for instance by removing the `<actuator>` data field and adding relevant sensors.
- I updated the inertias and masses based on the new values of [unitree_ros](https://github.com/unitreerobotics/unitree_ros/blob/master/robots/h1_description/urdf/h1.urdf) updated 4 months ago. To do so I converted the urdf to xml and copy-pasted the `<inertia>` data fields. This reduces the sim-to-real gap. For some reason most H1 descriptions you can find online are wrong. The Unitree website announces a weight of 47 kg. From memory the old mujoco menagerie xml was around 51 kg. I put the H1 I work with on a scale and its actual weight is 58.4 kg (stock H1 v1 without modifications), so most people train with ~20% missing mass. The total weight of a H1 is now 59.338kg based on unitree_ros information (so a 1 kg difference with what I measured, but at least it's better than before).
- I updated `test_asset_zoo.py` and `test_viser_conversions.py` to include H1, and added a new `test_h1_constants.py`.
- Data about H1's actuator M107 are unknown so I used values of G1_ACTUATOR_7520_14 as a placeholder to have an approximate value for the armature. I've set damping and stiffness values that I have seen in the literature and used for deployments myself.
- Based on how you did it for the `i2rt_yam`, I have added D435 camera and Mid360 lidar positions with information from `h1_with_hand.urdf` of [unitree_ros](https://github.com/unitreerobotics/unitree_ros/blob/master/robots/h1_description/urdf/h1.urdf), converted into xml. `sensorsize` and `focal` have been computed by asking an LLM because I wasn't sure about what the .xml was expecting for the D435. The `lidar_mid360` frame does nothing but could be used to attach a `RayCastSensorCfg` or `PinholeCameraPattern`

As validation, I launched a `uv run train Mjlab-Velocity-Flat-Unitree-H1 --env.scene.num-envs 4096` with default task parameters and it quickly converged to a basic locomotion behaviour after a few hundred epochs, like `uv run play Mjlab-Velocity-Flat-Unitree-H1 --checkpoint_file logs/rsl_rl/h1_velocity/2026-04-30_18-01-07/model_1300.pt`.

`make test-fast` all passed, including new tests.


kevinzakka:
Hi @paLeziart, I appreciate the effort but as mentioned in multiple previous issues, we're trying to keep the repo lean and adding an additional humanoid has marginal value. Hope you understand!

PR #966, opened 2026-05-02
per world material

omarrayyann:
Extends per-world variants so each variant can have its own material/texture.  `build_merged_variant_spec` copies them (like the mesh copy), and `build_variant_model` scatters them back per-world. Added tests with/without materials.

| Before | After |
|--------|-------|
|  |  |


kevinzakka (approved):
Thank you @omarrayyann, PR was basically perfect :)

PR #968, opened 2026-05-04
Remove use of deprecated warp-lang symbols

### LINKED ISSUE #967: Use of deprecated `warp` symbols breaks compatibility with 1.13
I was attempting to update to warp-lang 1.13.0, but ran into:

```
 File ".../mjlab/sim/sim.py", line 426, in _should_use_cuda_graph
    driver_ver = wp.context.runtime.driver_version
                 ^^^^^^^^^^
AttributeError: module 'warp' has no attribute 'context'. Did you mean: 'constant'?
```

This is because `warp.context` was previously deprecated and finally removed in 1.13 (see https://github.com/NVIDIA/warp/issues/1352). 

It should be sufficient to ensure that the tests pass without raising deprecation warnings with our current warp 1.12, at which point 1.13 should just work.

rdeits-bd:
Fixes #967 by switching to the non-deprecated warp API. This should unblock warp-lang 1.13+.

### DIFF
--- a/src/mjlab/sensor/raycast_sensor.py
+++ b/src/mjlab/sensor/raycast_sensor.py
@@ -435,7 +435,7 @@ class RayCastSensor(Sensor[RayCastData]):
     self._model: mjwarp.Model | None = None
     self._mj_model: mujoco.MjModel | None = None
     self._device: str | None = None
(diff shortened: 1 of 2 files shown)


kevinzakka (approved):
Thank you Robin!

PR #972, opened 2026-05-06
Fix pd_gains and effort_limits silently ignoring Operation objects

### LINKED ISSUE #971: `pd_gains` domain randomization silently no-ops for `operation: Operation`
Domain randomization events are expected to take `operation: str | dr.Operation`. However, the `pd_gains` event specifically expects only one of two literal strings: 

https://github.com/mujocolab/mjlab/blob/5a433e83fb27dc2b44b21863b47d3dae6e4dc211/src/mjlab/envs/mdp/dr/actuator.py#L78-L93

The more concerning issue is that there's no `else` clause there, so passing something like `dr.scale` instead of `"scale"` will result in the event being silently a no-op! At the very least we should error if we get an unsupported operation. Better yet would be to just correctly call the `operation` function if possible.

kevinzakka:
`dr.pd_gains` and `dr.effort_limits` accepted only `Literal["scale", "abs"]` strings for their `operation` parameter, unlike every other DR event which accepts `Operation | str`. Passing an `Operation` instance (e.g. `dr.scale` instead of `"scale"`) would silently skip all `if`/`elif` branches and do nothing.

Both functions now resolve the operation via `resolve_operation()`, compare against `op.name`, and raise `ValueError` for unsupported operations (e.g. `dr.add`). The `elif`-abs branches become `else` since the guard at the top already restricts to only scale/abs.

Fixes #971

### DIFF
--- a/src/mjlab/envs/mdp/dr/actuator.py
+++ b/src/mjlab/envs/mdp/dr/actuator.py
@@ -93,6 +93,7 @@ def pd_gains(
           default_biasprm[ctrl_ids, 2] * kd_samples
         )
       else:
+        assert op.name == "abs"
         env.sim.model.actuator_gainprm[env_ids[:, None], ctrl_ids, 0] = kp_samples
         env.sim.model.actuator_biasprm[env_ids[:, None], ctrl_ids, 1] = -kp_samples
         env.sim.model.actuator_biasprm[env_ids[:, None], ctrl_ids, 2] = -kd_samples
@@ -109,6 +110,7 @@ def pd_gains(
           kd=actuator.default_damping[env_ids] * kd_samples,
         )
       else:
+        assert op.name == "abs"
         actuator.set_gains(env_ids, kp=kp_samples, kd=kd_samples)
 
     else:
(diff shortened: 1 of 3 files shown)


src/mjlab/envs/mdp/dr/actuator.py
```diff
@@ -87,23 +92,23 @@ def pd_gains(
         env.sim.model.actuator_biasprm[env_ids[:, None], ctrl_ids, 2] = (
           default_biasprm[ctrl_ids, 2] * kd_samples
         )
-      elif operation == "abs":
+      else:
```
rdeits-bd, line 95: I'd suggest explicitly asserting `op.name == "abs"` here. Otherwise we're relying on the somewhat nonlocal check that `op.name in ("scale", "abs")` 50 lines earlier.

rdeits-bd (approved):
thanks!

rdeits-bd:
cc @bd-pdomanico

src/mjlab/envs/mdp/dr/actuator.py
kevinzakka: Good point! Should be done.

PR #975, opened 2026-05-07
Sample apply_body_impulse cooldown lazily on first step

### LINKED ISSUE #973: apply_body_impulse always fires on the first step, which doesn't match documentation
The docstring for apply_body_impulse says that the cooldown period is the first step in the lifecycle: 

https://github.com/mujocolab/mjlab/blob/80a12a5e12ffa66ec32a4b8167926cf4ba239fec/src/mjlab/envs/mdp/events.py#L367-L372

but actually the impulse will always fire on the very first step (and after any reset) due to the way `_interval_time_left ` is initialized. This is verified in the unit test: 

https://github.com/mujocolab/mjlab/blob/80a12a5e12ffa66ec32a4b8167926cf4ba239fec/tests/test_events.py#L593-L594

Is that the desired behavior? If so, we should at least reorder the documented lifecycle to match.

kevinzakka:
The `apply_body_impulse` docstring describes the lifecycle as starting with a cooldown, but `_interval_time_left` was initialized to zero, so an impulse always fired on the very first step — and on the first step after every reset. That made impulse timing perfectly correlated with episode resets, which is exactly the kind of artifact an RL policy can learn around.

The fix is to sample the initial cooldown lazily on the first `__call__` (and on the first call after each reset), since `cooldown_s` is only available at `__call__` time. A small `_needs_init` bool tensor tracks which envs need a fresh cooldown sample. The rest of the cycle is unchanged.

Updated `test_apply_body_impulse_basic` to skip the initial cooldown explicitly — that test exercises the trigger/sustain/expire cycle, not the cooldown — and added `test_apply_body_impulse_initial_cooldown` as a targeted regression test covering all three properties: no fire on the first calls, fires once cooldown elapses, and reset re-enters cooldown.

Fixes #973.

Thanks to @rdeits-bd for the catch.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -80,6 +80,10 @@ Changed
 Fixed
 ^^^^^
 
+- Fixed ``apply_body_impulse`` firing an impulse on the very first step (and
+  the first step after every reset) instead of starting with a cooldown as
+  documented. The cooldown is now sampled lazily on the first call so impulse
+  timing is decorrelated from episode resets (:issue:`973`).
 - Fixed ``dr.pd_gains`` and ``dr.effort_limits`` silently no-oping when
   passed an ``Operation`` object (e.g. ``dr.scale``) instead of a string.
   Both functions now accept ``Operation | str`` like every other DR event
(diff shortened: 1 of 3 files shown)


rdeits-bd:
I'm AFK until the morning, but thanks for the fix! Could you grab the
cooldown from self.cfg.params at init instead of needing to defer that
action to the first call?

PR #980, opened 2026-05-07
Adversarial motion prior implementation

ManuelActisCassa:
To launch **training** with Adversarial Motion Prior on G1 (training environment is provided) : 

`uv run train Mjlab-AMP-Flat-Unitree-G1 --motion-file-amp {filename}.csv --amp-weight {value} --resample {source_FPS} --{other options}`

**(required)** motion file amp csv file containing motion data (each row is observation for a single frame)

**(optionnal)** amp weight is 1.0 by default

**(optionnal)** resample uses linear interpolation to resample motion data from source fps to training step (automatically calculated). If not specified, no resampling is done.

**Important** : Observations in csv should match with observations specified in amp environment (joint_position, joint_velocity, root_linear_velocity ...).

To launch **inference** it is done as usual : 

`un run play Mjlab-AMP-Flat-Unitree-G1 --checkpoint-file {checkpoint path}`

In a few words, an amp environment describes high level objectives (velocity, stability, ...) while using a discriminator to reward motion style using a reference motion.

_____________

May need to add discriminator hyperparameters to CLI

_____________

Implementation based on paper : 
https://arxiv.org/abs/2104.02180
"AMP: Adversarial Motion Priors for Stylized Physics-Based Character Control"
By : Xue Bin Peng, Ze Ma, Pieter Abbeel, Sergey Levine, Angjoo Kanazawa

_____________

**Results so far :** 

Trained Unitree G1 with running motion from [LAFAN1 dataset](https://huggingface.co/datasets/lvhaidong/LAFAN1_Retargeting_Dataset) 

https://github.com/user-attachments/assets/44e9e75f-e35d-4162-971a-811409d2225b

Only high level objectives and discriminator for motion (few seconds of motions samples), no gait rewards here


louislelay:
Hey @ManuelActisCassa, really cool work! But TBH I'm not fully convinced it should be part of mjlab scope to natively support AMP, IMO mjlab's role should be about being flexible enough so people can implement AMP on their side

But super curious about @kevinzakka's opinion!

ManuelActisCassa:
Hey @louislelay! I already made an implementation in pal_mjlab that I’m currently testing on the kangaroo, but I thought it could still be worth opening a PR here. I tried to keep it as modular as possible.

louislelay:
Nice, can't wait to see videos of deployment ;)

songhat:
@ManuelActisCassa Hi,I gave the training a try, but it feels like it's learning pretty slowly. I was wondering what parameters I should pay attention to, and how long your training usually takes?

ManuelActisCassa:
Hi @songhat , thank you for giving it a try. The thing is training is extremely sensitive to amp term weight and discriminator hyper parameters. I made it so it logs into the metrics discriminator output, at the very least you can check that it doesn't immediately converge to -1. 

If it does, you can try tuning the discriminator so it is less aggressive (make it worse at telling apart motion and sim data).
It it doesn't, the signal should be enough for training, and you just need to find the right weight for amp reward.

As for training duration, I kept the default 30k iterations and it works ok, but it may converge to solution before or after depending on tuning.

I hope this answers your question.

songhat:
> Hi @songhat , thank you for giving it a try. The thing is training is extremely sensitive to amp term weight and discriminator hyper parameters. I made it so it logs into the metrics discriminator output, at the very least you can check that it doesn't immediately converge to -1.
> 
> If it does, you can try tuning the discriminator so it is less aggressive (make it worse at telling apart motion and sim data). It it doesn't, the signal should be enough for training, and you just need to find the right weight for amp reward.
> 
> As for training duration, I kept the default 30k iterations and it works ok, but it may converge to solution before or after depending on tuning.
> 
> I hope this answers your question.

Thanks for your reply, I'll give it a try.

kevinzakka:
Sorry I took so long to finally answer: this is awesome but indeed I don't think it should be part of mjlab core. We really want to keep this lean and efficient. Thanks for showcasing though @ManuelActisCassa!

PR #988, opened 2026-05-11
Fix multi-node distributed training using duplicate seeds across nodes

bd-pdomanico:
The per-process seed offset in `scripts/train.py` is derived from `LOCAL_RANK`, which is each process's rank within its node and only matches the global `RANK` on single-node runs. In any multi-node configuration the same `LOCAL_RANK` repeats on every node, so the seed offset collides and processes end up training with duplicate seeds. Concretely, training across two nodes with one GPU each gives both nodes `local_rank=0`, producing identical seeds on both nodes — defeating the "diversity in different processes" intent of the offset. Switching the offset to the global `RANK` makes every process across every node receive a unique seed.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -80,6 +80,9 @@ Changed
 Fixed
 ^^^^^
 
+- Fixed duplicate random seeds across nodes in multi-node training. The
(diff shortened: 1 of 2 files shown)


kevinzakka (approved):
Thank you @bd-pdomanico!

PR #989, opened 2026-05-14
Enforce @abstractmethod on ActionTerm and CommandTerm

bd-pdomanico:
The `@abc.abstractmethod` decorators on `ActionTerm` and `CommandTerm` were silently ignored because they did not inherit from `abc.ABC`, so the `ABCMeta` metaclass that makes the decorator effective was never installed. Subclasses that omitted a required override would construct successfully and only fail later at the call site with `NotImplementedError`.

Adding `abc.ABC` as a mixin on each class restores the intended behavior: incomplete subclasses now raise `TypeError` on instantiation, surfacing the mistake immediately. This is technically a behavior change — any subclass that was relying on the silent acceptance will now fail to construct.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -80,6 +80,10 @@ Changed
 Fixed
 ^^^^^
 
+- Fixed ``@abc.abstractmethod`` decorators being silently ignored on
+  ``ActionTerm`` and ``CommandTerm``. Subclasses that omit a required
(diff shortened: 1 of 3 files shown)


bd-pdomanico:
Note that this may raise exceptions for code that previously ran since some of these abstractmethods are not strictly required to run training. The alternative solution would be to remove the decorator and rely on the `NotImplementedError` which would be consistent with the current behavior without a misleading decorator.

PR #996, opened 2026-05-20
Fix rough-terrain video flicker from mutating env origins

### LINKED ISSUE #979: Offscreen training videos flicker heavily on rough-terrain tasks
When using the official `train` script with `--video True`, the recorded training videos on rough-terrain tasks flicker heavily.

```bash
train Mjlab-Velocity-Rough-Unitree-Go1 \
  --gpu-ids [0] \
  --env.scene.num-envs 4096 \
  --video True \
  --agent.max-iterations 10000
```

https://github.com/user-attachments/assets/85da0746-c887-4fce-b214-2a7f3d9a2a98

We can see that, the main robot view is unstable, with neighboring robots appearing and disappearing abruptly, making the video look as if it is switching context between nearby environments.

https://github.com/user-attachments/assets/c9a55998-09b7-46d7-9f15-52e7dea71954

Strangely, everything appears normal in the `Mjlab-Velocity-Flat-Unitree-Go1` task.

Environment:

- mjlab: 1.3.0

kevinzakka:
Offscreen training videos flicker heavily on rough-terrain tasks (#979). The renderer draws a few `env_origins`-nearest "neighbor" robots around the tracked env for context, but it recomputed that set every frame. On rough terrain the `terrain_levels` curriculum reassigns `env_origins` on reset, so the nearest set kept changing and the context robots popped in and out. Flat tasks looked fine because they disable the terrain curriculum, leaving origins static.

The fix computes the neighbor set once and caches it. `env_idx` and `nworld` are fixed for a recording session, so the context robots stay stable across the whole video. One tradeoff: under an active curriculum the cached neighbors are picked from the origins at the first rendered frame and may drift to other tiles over a long run, but a stable set is the right call for a focused tracking video.

Fixes #979

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -83,6 +83,11 @@ Changed
 Fixed
 ^^^^^
 
+- Fixed heavy flicker in offscreen training videos on rough-terrain tasks.
+  The renderer recomputed its context "neighbor" robots every frame from
+  ``env_origins``, which the terrain curriculum mutates on reset, so the
+  neighbor set kept changing and robots popped in and out. The neighbor
+  set is now computed once and cached (:issue:`979`).
 - Fixed duplicate random seeds across nodes in multi-node training. The
   per-process seed offset in ``scripts/train.py`` now uses the global
   ``RANK`` instead of ``LOCAL_RANK``. Contribution by @bd-pdomanico.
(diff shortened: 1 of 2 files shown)


kevinzakka:
@Zionshang could you let me know if this fixes your issue?

PR #997, opened 2026-05-21
Fix BadPixmap X11 error on viewer shutdown

saikishor:
Pressing Ctrl+C while the viewer is paused occasionally produces a BadPixmap X11 error and exits abnormally with code 1 instead of shutting down cleanly. The default X11 error handler calls exit(1) when it receives the error, bypassing Python's normal cleanup.

```
  Ctrl+C received. Shutting down viewer...                                                                                                                                              
  X Error of failed request:  BadPixmap (invalid Pixmap parameter)                                                                                                                      
    Major opcode of failed request:  148 ()                                                                                                                                             
    Minor opcode of failed request:  1                                                                                                                                                  
    Resource id in failed request:  0x6c0000f                                                                                                                                           
    Serial number of failed request:  1551                                                                                                                                              
    Current serial number in output stream:  1558   
```
  
This PR tries to address the issue by capturing the X11 error exceptions


kevinzakka:
@claude

kevinzakka:
Thanks for digging into this @saikishor! Since it's a cosmetic shutdown-only race, I'd rather not carry a global X11 error-handler override for it. Going to close for now.

saikishor:
> Thanks for digging into this @saikishor! Since it's a cosmetic shutdown-only race, I'd rather not carry a global X11 error-handler override for it. Going to close for now.

Agreed! Thanks for taking a look :)

PR #1003, opened 2026-05-24
Add sensor-based projected gravity for IMU pose randomization

kevinzakka:
Base linear/angular velocity are read from sensors on the IMU site, so they already reflect that site's pose. Projected gravity didn't: `mdp.projected_gravity` uses the root body orientation, so it can't observe IMU mounting offset.

This adds a `framezaxis` sensor `imu_upvector` (world Z expressed in the IMU site frame — the body-frame "up" vector) to Go1 and G1, and a `projected_gravity_from_sensor` observation that negates it to point along gravity. Because it reads the site frame, randomizing the site pose (e.g. via `dr.site_quat`) flows into the observation, unlike the entity-data version.

This is intentionally scoped to the reusable, behavior-neutral piece — the observation, the sensor, and the test. It doesn't change any task's observations or training behavior; wiring it into the locomotion configs with tuned IMU-randomization ranges can be a separate change.

Tests check the sensor and entity gravity against an independent quaternion→matrix ground truth, that the observation tracks a tilted site while the entity version stays put, and that the full `dr.site_quat` path perturbs the reading per-environment.

### DIFF
--- a/src/mjlab/envs/mdp/observations.py
+++ b/src/mjlab/envs/mdp/observations.py
@@ -106,6 +106,23 @@ def builtin_sensor(env: ManagerBasedRlEnv, sensor_name: str) -> torch.Tensor:
   return sensor.data
 
 
+def projected_gravity_from_sensor(
+  env: ManagerBasedRlEnv, sensor_name: str
+) -> torch.Tensor:
+  """Projected gravity from a ``framezaxis`` up-vector sensor.
+
+  The sensor is expected to output the world Z-axis expressed in the sensor's frame
+  (e.g. ``framezaxis`` with ``objtype=body objname=world`` and ``reftype=site``). That
+  is the body-frame "up" vector, so it is negated to point along gravity.
+
+  Unlike :func:`projected_gravity`, which uses the root body orientation, this reads
+  the sensor's site frame and therefore reflects IMU site pose randomization.
+  """
+  sensor = env.scene[sensor_name]
+  assert isinstance(sensor, BuiltinSensor)
+  return -sensor.data
+
+
 def height_scan(
   env: ManagerBasedRlEnv,
   sensor_name: str,
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -9,13 +9,11 @@ Added
 ^^^^^
 
 - Added ``mdp.projected_gravity_from_sensor``, an observation that derives
-  projected gravity from a ``framezaxis`` up-vector sensor on the IMU site
-  (negated) rather than from the root body orientation. Unlike
-  ``mdp.projected_gravity``, it reflects IMU site pose, so it can observe IMU
-  mounting domain randomization. Go1 and G1 now ship an ``imu_upvector``
-  sensor; the velocity tasks use this observation and randomize the IMU site
-  position/orientation via ``dr.site_pos`` / ``dr.site_quat``, and the G1
-  tracking config adds the same randomization.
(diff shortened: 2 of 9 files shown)


src/mjlab/envs/mdp/observations.py
```diff
@@ -106,6 +106,23 @@ def builtin_sensor(env: ManagerBasedRlEnv, sensor_name: str) -> torch.Tensor:
...
+  (e.g. ``framezaxis`` with ``objtype=body objname=world`` and ``reftype=site``). That
+  is the body-frame "up" vector, so it is negated to point along gravity.
+
+  Unlike :func:`projected_gravity`, which uses the root body orientation, this reads
+  the sensor's site frame and therefore reflects IMU site pose randomization.
+  """
+  sensor = env.scene[sensor_name]
+  assert isinstance(sensor, BuiltinSensor)
+  return -sensor.data
+
```
saikishor, line 124: @kevinzakka isn't it better to calculate it from the sensor orientation reading itself?, because we add noise to the IMU sensor readings, that way it can take into account both of them right?

kevinzakka: Noise in mjlab is applied per-term on the ground-truth state, so even if I derived gravity from a framequat reading, it wouldn't share that reading's noise. Each term is computed from clean state and corrupted independently. The shared IMU error you're describing is already modeled by the site_pos/site_quat DR: it perturbs the single IMU site, so the velocimeter, gyro, and framezaxis gravity all change together consistently. The per-term white noise on top stays independent which is actually what we want, since ang_vel and gravity are different physical quantities with independent measurement jitter. So framezaxis is enough here. Correlated white-noise modeling would be a separate broader change.

kevinzakka:
@claude

src/mjlab/envs/mdp/observations.py
saikishor: Understood. Thanks for the explanation :)

louislelay:
OOC @kevinzakka, why not a framequat sensor?

kevinzakka:
@louislelay framequat would need an extra rotation to pull out the up-vector, framezaxis gives me exactly what I want :)

louislelay:
Oh ok, super nice! I'll change my framequat sensors then ahah

qiayuanl:
Do you think a `projected_gravity_at_site` is better?


```
def projected_gravity_at_site(
  env: ManagerBasedRlEnv,
  site_name: str,
  asset_cfg: SceneEntityCfg,
) -> torch.Tensor:
  """Gravity projected into a named site's frame.

  Reads ``asset.data.site_quat_w`` (auto-rotated by ``dr.site_quat``) and
  projects world-down into the site frame. Matches what an IMU mounted at
  the site reads, so randomizing the site quat shifts both this term and
  the gyro-based ``base_ang_vel`` consistently.
  """
  asset = env.scene[asset_cfg.name]
  site_idx = asset.site_names.index(site_name)
  site_quat_w = asset.data.site_quat_w[:, site_idx]
  world_down = torch.zeros(env.num_envs, 3, device=env.device)
  world_down[..., 2] = -1.0
  return quat_apply_inverse(site_quat_w, world_down)
```

kevinzakka:
@qiayuanl That is indeed an equivalent no-XML alternative. Note both it and framezaxis read from site_xmat so it has identical DR/noise behavior. I'm going to keep the current version since it keeps every IMU channel (lin/ang vel + gravity) on a single MuJoCo sensor path (and leaves the door open to MuJoCo-level sensor noise/cutoff later).

PR #1007, opened 2026-05-24
Fix tracking MPKPE metric to use the global reference

### LINKED ISSUE #1006: Potential wrong tracking task MPKPE metric computation
Hi @kevinzakka, @qiayuanl and the development team, thank you for making this great project! Though I spotted a potential wrong metric computation in the following snippets, feel free to close this issue if the following observation might be wrong.

### A

For global world frame meansurement, there is `compute_mpkpe`:

https://github.com/mujocolab/mjlab/blob/de120a9a464373118d07bb1946c92f5b23f4ce96/src/mjlab/tasks/tracking/mdp/metrics.py#L13-L21

It uses `command.body_pos_relative_w` to compute the error with `command.robot_body_pos_w`

### B

For anchor-aligned relative frame measurement, there is `compute_root_relative_mpkpe`:

https://github.com/mujocolab/mjlab/blob/de120a9a464373118d07bb1946c92f5b23f4ce96/src/mjlab/tasks/tracking/mdp/metrics.py#L21-L43

It uses `(command.body_pos_w - ref_anchor_pos)` to compute the error with `(command.robot_body_pos_w - robot_anchor_pos)`

### Potential issue

Since `command.body_pos_w` is the single source of the ground truth under the world frame, I think `A` should use `command.body_pos_w` as well instead of the relative quantity `command.body_pos_relative_w`, which is a per-step derived variable used in Beyondmimic to produce reward signals.

### Proposed fix

Change the following line to `pos_error = command.body_pos_w - command.robot_body_pos_w`:

https://github.com/mujocolab/mjlab/blob/de120a9a464373118d07bb1946c92f5b23f4ce96/src/mjlab/tasks/tracking/mdp/metrics.py#L19

kevinzakka:
compute_mpkpe compared the robot against body_pos_relative_w, which is the reference re-anchored to the robot's root every step (the quantity the tracking reward optimizes). That cancels global drift, so the metric never measured the global-frame error it documents and nearly duplicated compute_root_relative_mpkpe. It now uses the true global reference body_pos_w. Adds a test pinning this.

Thanks to @zhuoqun-chen for reporting.

Fixes #1006.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -97,6 +97,12 @@ Changed
 Fixed
 ^^^^^
 
+- Fixed the tracking MPKPE metric (``compute_mpkpe``) using the reward's
+  drift-cancelled reference ``body_pos_relative_w`` instead of the true
+  global reference ``body_pos_w``. It is documented to measure global-frame
+  keypoint error but previously never captured global drift and nearly
+  duplicated the root-relative metric. It now uses ``body_pos_w``
+  (:issue:`1006`).
 - Fixed heavy flicker in offscreen training videos on rough-terrain tasks.
   The renderer recomputed its context "neighbor" robots every frame from
   ``env_origins``, which the terrain curriculum mutates on reset, so the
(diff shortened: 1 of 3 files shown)


kevinzakka:
@claude

PR #1013, opened 2026-05-25
Reject entity specs with multiple freejoints

kevinzakka:
An Entity has at most one freejoint (its root). A second freejoint was previously accepted silently and only surfaced later as a cryptic shape mismatch when writing root state. This raises a clear error at construction instead, pointing users to model each detached floating body as its own `SceneCfg.entities` entry.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -54,6 +54,12 @@ Added
 Changed
 ^^^^^^^
 
+- ``Entity`` now raises a clear error at construction when its spec contains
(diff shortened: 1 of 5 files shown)


kevinzakka:
@claude

kevinzakka:
MuJoCo's compiler rejects a freejoint declared inside a sub-body, so it can't be "silently missed" at runtime. See [src/user/user_model.cc](https://github.com/google-deepmind/mujoco/blob/main/src/user/user_model.cc#L2756-L2758).

PR #1014, opened 2026-05-26
Add BuiltinPdActuator: implicit-integration PD with position and velocity references

kevinzakka:
Adds `BuiltinPdActuator`: the implicit-integration counterpart to `IdealPdActuator`. Same interface (position + velocity targets, kp/kd), but the PD is expressed as native MuJoCo `<position>` and `<velocity>` elements instead of a Python-computed force on a `<motor>`. The `implicit` / `implicitfast` integrators then include the kp/kd derivatives in their velocity update, so the actuator stays stable at gain/timestep combinations where explicit Python PD would diverge. Useful when you want to run a real motor's stiff on-board PD gains in sim.

`effort_limit` is a sum-clamp on `kp*err + kd*verr` via `jnt_actuatorfrcrange` (or `tendon_actuatorfrcrange`), matching the single torque limit a real motor controller has. `dr.pd_gains` and `dr.effort_limits` both branch on the new actuator.

### DIFF
--- a/src/mjlab/actuator/builtin_actuator.py
+++ b/src/mjlab/actuator/builtin_actuator.py
@@ -10,6 +10,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 import mujoco
+import numpy as np
 import torch
 
 from mjlab.actuator.actuator import (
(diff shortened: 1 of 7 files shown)


kevinzakka:
@claude

src/mjlab/actuator/builtin_actuator.py
```diff
@@ -91,6 +92,99 @@ def compute(self, cmd: ActuatorCmd) -> torch.Tensor:
...
+    self,
+    cfg: BuiltinPdActuatorCfg,
+    entity: Entity,
+    target_ids: list[int],
+    target_names: list[str],
+  ) -> None:
+    super().__init__(cfg, entity, target_ids, target_names)
+
+  @property
+  def half(self) -> int:
```
rdeits-bd: This function name was pretty surprising. `num_targets` might be more obvious.

tests/test_builtin_pd_actuator.py
```diff
@@ -0,0 +1,441 @@
...
+      target_names_expr=("x",),
+      stiffness=1.0,
+      damping=1.0,
+      transmission_type=TransmissionType.SITE,
+    )
+
+
+def test_armature_applied_once(device):
+  """Joint armature must come from the position element only; double-applying
+  would silently double dof_armature."""
```
rdeits-bd, line 111: Thanks for testing this. I was also concerned we'd see this kind of issue

rdeits-bd (approved):
Looks really nice, thank you!

src/mjlab/actuator/builtin_actuator.py
kevinzakka: Agreed, renamed!

kevinzakka:
@rdeits-bd Addressed your PR comments (and Claude's).

PR #1016, opened 2026-05-27
Add BuiltinDcMotorActuator: native MuJoCo <dcmotor> wrapper

kevinzakka:
Adds `BuiltinDcMotorActuator`, which wraps the newly introduced antive MuJoCo [`<dcmotor>`](https://mujoco.readthedocs.io/en/stable/XMLreference.html#actuator-dcmotor) actuator.

### DIFF
--- a/src/mjlab/actuator/builtin_actuator.py
+++ b/src/mjlab/actuator/builtin_actuator.py
@@ -7,7 +7,8 @@ created programmatically via the MjSpec API.
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Literal
+from enum import IntEnum
+from typing import TYPE_CHECKING
 
 import mujoco
 import numpy as np
@@ -95,17 +96,17 @@ class BuiltinPositionActuator(Actuator[BuiltinPositionActuatorCfg]):
 
 @dataclass(kw_only=True)
 class BuiltinPdActuatorCfg(ActuatorCfg):
-  """Implicit-integration version of ``IdealPdActuator``.
+  """Implicit-integration version of IdealPdActuator.
 
-  Both consume a position target and a velocity target with kp/kd gains.
-  The difference is in how the PD is delivered to MuJoCo: ``IdealPdActuator`` computes
-  the PD force in Python and feeds it to a ``<motor>`` element, which MuJoCo sees as an
-  opaque external force. This actuator expresses the PD as native MuJoCo elements (a
-  ``<position>`` carrying kp, a ``<velocity>`` carrying kd), so the ``implicit`` and
-  ``implicitfast`` integrators include the kp/kd derivatives in their velocity update.
-  That makes the actuator numerically stable at gain/timestep combinations where
-  explicit Python PD would diverge, which matters when you want to run a real motor's
-  stiff on-board PD gains in sim.
+  Both consume a position target and a velocity target with kp/kd gains. The
+  difference is in how the PD is delivered to MuJoCo: IdealPdActuator computes
+  the PD force in Python and feeds it to a ``<motor>`` element, which MuJoCo
+  sees as an opaque external force. This actuator expresses the PD as native
+  MuJoCo elements (a ``<position>`` carrying kp, a ``<velocity>`` carrying kd),
+  so the implicit and implicitfast integrators include the kp/kd derivatives
+  in their velocity update. That makes the actuator numerically stable at
+  gain/timestep combinations where explicit Python PD would diverge, which
+  matters when you want to run a real motor's stiff on-board PD gains in sim.
   """
 
   stiffness: float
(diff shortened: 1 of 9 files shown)


kevinzakka:
@rdeits-bd Would appreciate a review 🙏

bd-aadubredu:
@bd-xwu

src/mjlab/actuator/builtin_actuator.py
```diff
@@ -239,6 +240,254 @@ def compute(self, cmd: ActuatorCmd) -> torch.Tensor:
...
+  * ``nominal=(V_n, tau_stall, omega_no_load)``: datasheet form.
+  * ``motor_const=(Kt, Ke)`` + ``resistance=R``: physical form.
+
+  ``mjs_setToDCMotor`` derives ``K`` and ``R`` (including the viscous-damping
+  correction) and packs the generic ``gainprm`` / ``biasprm`` / ``dynprm``
+  slots.
+
+  Optional extensions, off by default: ``integral_gain`` / ``integral_limit``,
+  ``slew_rate``, ``inductance`` / ``electrical_time_constant``, ``thermal``,
+  ``lugre``, ``cogging``. ``dr.pd_gains`` randomizes only kp and kd; for DR
```
rdeits-bd: Nit: The ``dr.pd_gains`` section should probably be its own paragraph.

src/mjlab/actuator/builtin_actuator.py
```diff
@@ -239,6 +240,254 @@ def compute(self, cmd: ActuatorCmd) -> torch.Tensor:
     return cmd.effort_target
 
 
+DcMotorInputMode = Literal["voltage", "position", "velocity"]
+# Values match MuJoCo's <dcmotor input=...> enum, consumed by mjs_setToDCMotor
+# and read as gainprm[8] (see mujoco/src/xml/xml_native_reader.cc dcmotorinput_map).
+_DCMOTOR_INPUT_MODE: dict[DcMotorInputMode, int] = {
```
rdeits-bd: Can this be an enum instead of a `Literal` and a `dict[str, int]`?

src/mjlab/actuator/builtin_actuator.py
```diff
@@ -239,6 +240,254 @@ def compute(self, cmd: ActuatorCmd) -> torch.Tensor:
...
+  velocity-dependent torque clamp on top of a ``<motor>`` element; this one
+  is the real motor model with back-EMF and an optional native PID.
+  """
+
+  mode: DcMotorInputMode = "position"
+  """``ctrl`` input semantics. See class docstring."""
+
+  nominal: tuple[float, float, float] | None = None
+  """Datasheet characterization: ``(nominal_voltage, stall_torque,
+  no_load_speed)``. Provide this XOR ``motor_const`` + ``resistance``."""
```
rdeits-bd: Alternatively, something like:

```
@dataclass
class DcMotorDatasheetParams:
  nominal_voltage: float
  ...

@dataclass
class DcMotorPhysicalParams:
  motor_const: tuple[float, float]
  ...

...
class BuiltinDcMotorActuatorCfg:
  ...
  motor_params: DcMotorDatasheetParams | DcMotorPhsyicalParams
```

Would be more explicit, and enforces the either/or in an obvious way with no opportunity for users to specify more or less than what is required. OTOH it moves a bit further from the exact mujoco API.

rdeits-bd (approved):
Very cool! 

Thinking about this some more, it's not obvious to me if it would be possible to express the nonzero velocity reference handling from https://github.com/mujocolab/mjlab/pull/1014 for a dcmotor. What would that even look like? The trick of creating separate position and velocity actuators won't work because the electrical effects need to be considered simultaneously. This might be more of a core mujoco issue.

src/mjlab/actuator/builtin_actuator.py
kevinzakka: Done.

src/mjlab/actuator/builtin_actuator.py
kevinzakka: Yeah that's much cleaner!

src/mjlab/actuator/builtin_actuator.py
kevinzakka: Agreed and implemented!

kevinzakka:
Thanks for the thorough review @rdeits-bd! The BuiltinPdActuator trick doesn't carry over due to a MuJoCo-level limitation: 2 <dcmotor> elems would double count back-emf and clamp Vmax independently. POSITION mode also hard-codes `kd * (-qdot)`. Probably worth filing upstream (cc @yuvaltassa).

rdeits-bd:
Yeah, it seems like my only option would be to switch to an explicit actuator wrapping a `<dcmotor>` in voltage control mode, with the actual PD law implemented at sim rate. 

Presumably I'd need to make a new actuator type for this. Or, actually, should the mjlab DcMotorActuator just become that?

kevinzakka:
Gonna merge this PR, let's discuss this in a new issue I'll create!

Issue #1019, opened 2026-05-27
Should SceneEntityCfg ID collections become on-device tensors?


rdeits-bd:
Currently, when we resolve a `SceneEntityCfg` with a set of body or joint or site names, we end up with `body_ids`, `joint_ids`, etc. as a Python `list[int]`. We typically then do something like:

```py
# Grab the positions of the joints I care about
entity.data.joint_pos[:, entity_cfg.joint_ids]
```

Because `entity_cfg.joint_ids` is a Python list, we necessarily hit this implicit CUDA synchronization: https://docs.nvidia.com/dl-cuda-graph/torch-cuda-graph/sync-free-code.html#indexing-tensors (specifically the `x_gpu[idx_list]  # cuStreamSynchronize, implicit blocking .to()` case). The `joint_ids` list needs to be moved into a brand-new tensor on the GPU every time we hit that line. This creates unnecessary allocations and synchronization steps. 

We've been working around this by ditching the functional reward/observation format and caching `self._joint_ids: Tensor` in a reward/observation class. But it would be pretty nice not to have to do that. 

Is there some performance trap that we're avoiding by using a plain `list[int]` here? Or could we switch this to tensors and get rid of these implicit CUDA syncs in common cases? 

cc @bd-pdomanico who pointed this out originally. 
`

PR #1020, opened 2026-05-27
Materialize SceneEntityCfg ids as device tensors to avoid per-step CUDA syncs

kevinzakka:
Prompted by #1019, prototyping using torch tensors for `*_ids` to avoid per-step CUDA syncs from `tensor[:, list]` indexing.

| Task | Baseline | This PR | Δ |
|---|---:|---:|---:|
| Velocity-Flat-Go1 | 307,824 | 329,060 | **+6.9%** |
| Tracking-Flat-G1 | 194,248 | 194,655 | +0.2% |
| Lift-Cube-Yam | 807,465 | 858,967 | **+6.4%** |

### DIFF
--- a/scripts/benchmarks/measure_synthetic_throughput.py
+++ b/scripts/benchmarks/measure_synthetic_throughput.py
@@ -1,0 +1,151 @@
+"""Stress-test SceneEntityCfg ID indexing with a synthetic many-term env.
+
+Builds a cartpole env inflated with N extra observation and reward terms, each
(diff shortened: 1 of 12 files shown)


kevinzakka:
cc @rdeits-bd

rdeits-bd:
Very cool. @bd-pdomanico is definitely more of an expert on this than I am. From what I've learned, the major benefit comes when we eliminate all the CUDA syncs, at which point the physics steps and MPD terms can start to interleave their execution where possible. Eliminating this class of syncs would be a necessary step in that direction.

PR #1028, opened 2026-05-29
Make curriculum terrain difficulty deterministic.

### LINKED ISSUE #1027: Curriculum terrain difficulty doesn't seem to match documentation
The docs for curriculum terrain difficulty say:

> In curriculum mode, difficulty is determined by the row: row 0 gets the minimum, row num_rows - 1 gets the maximum.

which makes it sound deterministic. But actually the difficulty is still randomly sampled here: 

https://github.com/mujocolab/mjlab/blob/dd973373a39d3ea687109303acad657eaf892d42/src/mjlab/terrains/terrain_generator.py#L263

This has a pretty surprising effect for a terrain grid with just 1 row. In that case, the range of possible difficulties for that single row is [0, 1], sampled once per column. But in curriculum mode, we only get one column per terrain type. That means that, with num_rows=1, each terrain type gets *just one* sampled difficulty, from [0, 1], chosen at startup. This makes for a pretty unpredictable training setup, and it makes the random seed have a much larger effect than one might imagine. 

This behavior isn't necessarily wrong, but I do think we should at least call it out in the docs.

Also, speaking of the docs, I think this section is backwards?: https://github.com/mujocolab/mjlab/blob/main/docs/source/terrain.rst#grid-layout

Image

That seems like the opposite of what happens here, where curriculum mode means one column per type: https://github.com/mujocolab/mjlab/blob/dd973373a39d3ea687109303acad657eaf892d42/src/mjlab/terrains/terrain_generator.py#L157-L159

kevinzakka:
In #1027, Robin correctly pointed out that curriculum-mode terrain difficulty is randomly sampled within each row's bin, which makes the `difficulty_range` endpoints unreachable and makes `num_rows=1` behave randomly instead of as the easiest level.

This PR addresses that by replacing the sampling with a deterministic interpolation: `lower + (upper - lower) * row / max(num_rows - 1, 1)`, so row 0 hits the configured min, row N-1 hits the max, and `num_rows=1` produces the easiest difficulty.

It also fixes the doc inaccuracies Robin flagged, namely the inverted description of curriculum-mode column allocation (in #811 it became one column per terrain type, with `proportion` controlling robot spawning rather than column count).

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -17,6 +17,9 @@ Changed
 ^^^^^^^
 
 - Bumped ``rsl-rl-lib`` from 5.2.0 to 5.4.0.
+- Curriculum-mode terrain difficulty is now deterministic across rows
+  and reaches the configured ``difficulty_range`` endpoints
+  (:issue:`1027`).
 
 Fixed
 ^^^^^
--- a/docs/source/terrain.rst
+++ b/docs/source/terrain.rst
@@ -49,7 +49,6 @@ the geometry and how it scales with difficulty.
         terrain_generator=TerrainGeneratorCfg(
             size=(8.0, 8.0),
             num_rows=10,
-            num_cols=20,
             border_width=20.0,
             curriculum=True,
             sub_terrains={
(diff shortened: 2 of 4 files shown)


kevinzakka:
@rdeits-bd This is going to be a breaking change but probably one for the better?

rdeits-bd (approved):
Looks good, and thanks for the doc updates!

PR #1029, opened 2026-05-29
Fix GPU selection crash with MIG devices

flferretti:
When NVIDIA MIG is enabled, `CUDA_VISIBLE_DEVICES` contains UUIDs like "MIG-GPU-abc-123" instead of numeric indices. The `int()` call in `select_gpus` would crash on these. This change parses numeric entries as int preserving existing behavior and passes non-numeric entries through as strings.
I also added a `GpuId = int | str` type alias to reflect this in the type system and a test covering MIG UUIDs.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -18,6 +18,11 @@ Changed
 
 - Bumped ``rsl-rl-lib`` from 5.2.0 to 5.4.0.
 
+Fixed
+^^^^^
+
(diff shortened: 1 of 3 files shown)


kevinzakka (changes requested):
Thank you @flferretti, I didn't even know this was a thing lol.

Can you just add a changelog entry?

flferretti:
> Thank you @flferretti, I didn't even know this was a thing lol.
> 
> Can you just add a changelog entry?

Done, thank you!

kevinzakka (approved):

PR #1030, opened 2026-05-29
Add an option to capture true terminal observations under auto-reset.

### LINKED ISSUE #1026: How to let environment get termination observations from extras?
Hi @kevinzakka and the team,

currently the manager-based environment doesn't log `termination_observations` into `extras["log"]`, is this a design choice or there is possibility to implement this feature?

The reason behind this is that for algorithms that comsume consequtive observation histories (especially newly transitioned observation)  to compute additional rewards like `AMP` algorithm, since in `rsl_rl` the algorithm class `process_env_step` by taking environment returned observation, for auto-reset env, the `obs` very likely returned new episode initial observation for `new_ids`. When computing `AMP` style reward, the final step task reward will wrongly blend with mismatched style reward returned by discriminator.

This is also useful for other algorithm libraries that reply on auto-reset env API but need to consume `termination_observations` as well.

kevinzakka:
Some algorithms need the true terminal state at episode boundaries: value bootstrapping at truncation needs the real next state to bootstrap from (the original ask in #900), and AMP-style rewards need it because the discriminator scores `(s_t, s_{t+1})` transitions, so a wrong final state means scoring a transition into the next episode's initial state instead of the actual one (#1026). Our current auto-reset returns the post-reset observation for done environments, so that terminal state is lost.

This PR adds an opt-in `capture_terminal_observations` flag. When it's on and an environment resets, `step()` stashes that environment's true terminal observation in `extras["terminal"]` before the reset. It's off by default and the key is only present on reset steps, so existing setups are unchanged.

The snapshot is read-only (it never touches the live history/delay buffers or the RNG), so the terminal observation is the noise-free terminal state with delay and history applied. To turn it into a full next observation:

```python
next_obs = obs[group].clone()
if (term := extras.get("terminal")) is not None:
    next_obs[term["env_ids"]] = term["observations"][group]
```

Fixes #1026

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -8,6 +8,14 @@ Upcoming version (not yet released)
 Added
 ^^^^^
 
+- Added ``ManagerBasedRlEnvCfg.capture_terminal_observations``. When enabled
+  (with ``auto_reset=True``), ``step()`` stores the true terminal observation
+  (the real next state, before reset) for done environments in
+  ``extras["terminal"]``, so algorithms that need the real next state at episode
+  boundaries (value bootstrapping at truncation, AMP-style style rewards) can
+  access it without disabling auto-reset. The snapshot is read-only: it applies
+  delay and history but not observation noise, and enabling the flag does not
+  change the training trajectory (:issue:`1026`, :issue:`900`).
 - Added ``BuiltinDcMotorActuator``, a native MuJoCo ``<dcmotor>`` wrapper.
   Supports voltage / position / velocity input modes with back-EMF,
   configurable motor constants, and optional integral, slew, inductance,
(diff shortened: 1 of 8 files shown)


kevinzakka:
cc @realquantumcookie

realquantumcookie:
Cool! Right now `auto_reset` already works for my case. Congrats on the 1.4.0 release btw :)

zhuoqun-chen:
Hi @realquantumcookie, before this PR is landed, for auto reset `env_ids`, the observation is actually the initial observations in the new episodes, so will this affect your use cases especially for computing $Q(s_{t+1}, a_{t+1})$?

realquantumcookie:
@zhuoqun-chen Before this PR lands, I disabled auto-reset so when one of the environment terminates it gives the correct $s_{t+1}$. You just have to manually reset the environment calling `reset(env_idx=...)` after any truncation / termination occurs

PR #1036, opened 2026-06-01
Set MUJOCO_GL default before mujoco import

rakhimovv:
mjlab was setting `MUJOCO_GL=egl` inside `launch_training` in `src/mjlab/scripts/train.py`, but the top-level imports there had already pulled in `mujoco.rendering.classic.gl_context`, which reads `MUJOCO_GL` once at import time. By then mujoco had picked its GL backend against an empty value and never `dlopen`'d a platform library, so `train --video True` crashed with `mujoco.FatalError: an OpenGL platform library has not been loaded` on hosts where `MUJOCO_GL` wasn't set in the shell. `play` and `demo` share the same import chain and hit the same crash whenever they render.

This moves the default into `src/mjlab/__init__.py` ahead of any other import in the package, so the env var is set before any code that might pull in mujoco. That fixes all three entry points in one place and also covers library users who never go through the CLI scripts.

`setdefault` (not plain assignment) preserves user overrides like `MUJOCO_GL=osmesa` for clusters without EGL. Verified on a headless H100 container: `unset MUJOCO_GL && uv run train Mjlab-Cartpole-Balance --video True ...` finishes and uploads videos to W&B, and `MUJOCO_GL=osmesa python -c 'import mjlab'` keeps `osmesa` set.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -60,6 +60,7 @@ Fixed
   threshold, where the stone grid re-tiled as its spacing crossed an integer
   boundary, and leaving an oversized gap around the center platform. The grid is
   now difficulty-independent and the platform snaps to it as a clean island.
+- Fixed ``train --video``, ``play``, and ``demo`` crashing with ``OpenGL platform library not loaded`` on headless hosts that don't pre-set ``MUJOCO_GL``. The default is now applied in ``mjlab/__init__.py`` so it takes effect before mujoco's GL backend selection runs.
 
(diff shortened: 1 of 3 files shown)


kevinzakka (approved):
Hey @rakhimovv, thanks for this. I had to push a small fix because the unconditional `setdefault("MUJOCO_GL", "egl")` breaks macOS since mujoco validates MUJOCO_GL against a platform-specific allowlist at import time and rejects egl on Darwin.

PR #1040, opened 2026-06-02
Add actuator velocity limit metadata

cmjang:
## Summary

This PR adds an optional `velocity_limit` field to the common `ActuatorCfg` base class.

The field is intended as actuator/joint velocity-limit metadata that downstream terms can read, for example reward terms that penalize joints above an actuator's recommended or hardware velocity limit. It is validated when provided, but it does not change runtime control behavior by itself.

## Motivation

Isaac Lab exposes actuator velocity limits as actuator configuration data, separate from simulation-side solver limits such as `velocity_limit_sim`:

- Reference: https://isaac-sim.github.io/IsaacLab/main/source/api/lab/isaaclab.actuators.html

For mjlab/MuJoCo, this PR keeps the change conservative: MuJoCo/MJWarp does not provide a direct Isaac Lab-style joint velocity solver cap through the same actuator config path, so this field is metadata only. It gives task/reward code a standard place to store the velocity limit without implying that the simulator will clamp joint velocity or actuator commands.

## Scope

Included:

- Add `ActuatorCfg.velocity_limit: float | None = None`.
- Validate that `velocity_limit` is positive when provided.
- Keep `DcMotorActuatorCfg` compatible with the new base field while preserving its existing runtime requirement that DC motor configs provide `velocity_limit`.
- Add actuator tests covering metadata acceptance and validation.
- Confirm that setting this metadata does not clip a velocity actuator command.

Not included in this PR:

- No reward implementation.
- No EntityData plumbing.
- No asset zoo config updates.
- No documentation updates.
- No MuJoCo model/compiler changes.

## Test Plan

- `uv run pyright`
- `uv run ruff check src/mjlab/actuator/actuator.py src/mjlab/actuator/dc_actuator.py tests/test_actuator.py tests/test_dc_actuator.py`
- `uv run python -m py_compile src/mjlab/actuator/actuator.py src/mjlab/actuator/dc_actuator.py tests/test_actuator.py tests/test_dc_actuator.py`
- `uv run ty check src/mjlab/actuator/actuator.py src/mjlab/actuator/dc_actuator.py tests/test_actuator.py tests/test_dc_actuator.py`
- `uv run pytest tests/test_actuator.py tests/test_dc_actuator.py`

### DIFF
--- a/src/mjlab/actuator/actuator.py
+++ b/src/mjlab/actuator/actuator.py
@@ -46,6 +46,15 @@ class ActuatorCfg(ABC):
   armature: float | None = None
   """Reflected rotor inertia. None preserves the XML value."""
 
+  velocity_limit: float | None = None
+  """Maximum joint velocity metadata. None leaves it unspecified.
+
+  MuJoCo does not expose an Isaac Lab-style solver velocity cap for joints or
+  tendons, so this common field is not written into the compiled model. Actuator
+  models may use it for their own torque-speed limits, and downstream terms can
+  read it as hardware metadata.
+  """
+
   frictionloss: float | None = None
   """Friction loss force limit. None preserves the XML value.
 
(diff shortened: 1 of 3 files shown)


kevinzakka:
@rdeits-bd I am hesitant about merging this but wondering what you think?

rdeits-bd:
Agreed. Supporting this field but not actually using it in the physics model seems like a trap.

XiaoPengYouCode:
I think what we really need is a complete system for detailed modeling of the motor TN characteristic curve, as mentioned in the [OmniXtreme paper](https://arxiv.org/abs/2602.23843). Simply describing the speed limit of the motor does not have universal value in practical applications.

louislelay:
+1 not sure it's a good idea

PR #1043, opened 2026-06-03
Fix per-axis domain randomization events clobbering each other

### LINKED ISSUE #1042: Multiple geom_size events with different `axes` clobber each other
It looks like having multiple `dr.geom_size` events which target different `axes` on the same `geom` result later events clobbering the effects of earlier events silently. For example, take this event cfg:

```py
{
    "foot_scale_x":
        EventTermCfg(
            mode="startup",
            func=dr.geom_size,
            params={
                "asset_cfg": SceneEntityCfg("robot", geom_names=("[lr]_foot")),
                "operation": dr.scale,
                "axes": [0],
                "ranges": (0.1, 0.1),
            },
        ),
    "foot_scale_y":
        EventTermCfg(
            mode="startup",
            func=dr.geom_size,
            params={
                "asset_cfg": SceneEntityCfg("robot", geom_names=("[lr]_foot")),
                "operation": dr.scale,
                "axes": [1],
                "ranges": (1.0, 1.0),
            },
        ),
}
```

The first event sets the scaling of axis 0 to 0.1, while the second event sets the scaling of axis 1 to 1.0. When I visualize the resulting environment, I see that the foot geom has been scaled by 1.0 in *both* dimensions. If I delete `foot_scale_y` then I *do* see the x-axis scaling from `foot_scale_x` applied. 

I think what I'm seeing is that the scaling applied by `foot_scale_y` is clobbering the effect of `foot_scale_x`. We should either error if this isn't supported or ensure that these events apply correctly together.

kevinzakka:
dr.geom_size and the other multi-axis randomization functions let you target a subset of axes via the axes argument. Two events targeting different axes of the same field don't compose: the second silently resets the axes the first set.

The shared engine writes the full indexed slice back to the model field every call. For scale and add, base values are the compile-time defaults and non-target axes carry the operation identity, so the write lands default * 1 (or default + 0) on every axis the event didn't target, overwriting any prior event's work.

This restricts the write to the targeted axes. Per-axis events now compose, across every multi-axis DR function.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -38,6 +38,10 @@ Changed
 Fixed
 ^^^^^
 
+- Fixed domain randomization events that target different ``axes`` of the same
+  model field (e.g. two ``dr.geom_size`` events scaling axis 0 and axis 1
+  separately) silently clobbering each other. Each event now writes back only
+  the axes it targeted, so per-axis events compose (:issue:`1042`).
 - Fixed ``select_gpus`` crashing when ``CUDA_VISIBLE_DEVICES`` contains MIG UUIDs instead of numeric indices.
 - Fixed pyramid-stairs terrains (``BoxPyramidStairsTerrainCfg``,
   ``BoxInvertedPyramidStairsTerrainCfg``, and ``BoxOpenStairsTerrainCfg``)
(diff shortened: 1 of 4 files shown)


kevinzakka:
@claude

rdeits-bd (approved):
Looks great, thanks!

PR #1045, opened 2026-06-04
Fix circular buffer push broadcast shape for non-2D data

ZzzzzzS:
CircularBuffer.push crashes with a broadcast shape mismatch when the buffered data has more than two dimensions. The bug surfaces in DcMotorActuator when delay is configured with a variable lag range (delay_min_lag != delay_max_lag) and the actuator operates in position mode. In this configuration, apply_delay stacks position_target, velocity_target, and effort_target into a 3-D tensor with shape (num_envs, num_targets, 3), then pushes it into the delay buffer.

The root cause is in the first-push backfill logic, which hard-coded [None, :, None] for the torch.where condition. This produces a condition of shape (1, batch_size, 1), but the buffer slice has shape (max_len, num_envs, num_targets, 3). During broadcasting, PyTorch misaligns the condition's batch dimension with the buffer's num_targets dimension. When those differ — for example num_envs=4 but the actuator controls 2 joints — the broadcast fails with a 4 vs 2 runtime error. The bug is latent when delay_min_lag == delay_max_lag because the fixed-delay path does not exercise this backfill code.

The fix replaces the hard-coded views with is_first_push.view(1, batch_size, *([1] * (data.ndim - 1))) and data.unsqueeze(0), so the broadcast shape is derived dynamically from data.ndim and works for tensors of any rank. A regression test for DC motor delay in position mode with delay_min_lag=2, delay_max_lag=4 is included to cover this path.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -68,6 +68,11 @@ Fixed
   platform library not loaded`` on headless Linux hosts that don't pre-set
   ``MUJOCO_GL``. The default is now applied in ``mjlab/__init__.py`` (Linux
   only) so it takes effect before mujoco's GL backend selection runs.
+- Fixed ``CircularBuffer.push`` crashing with a broadcast shape mismatch
+  when the buffered data has more than two dimensions. The first-push
+  backfill previously hard-coded ``[None, :, None]`` and ``[None, :, :]``
+  views, which only matched 2-D tensors; it now computes the broadcast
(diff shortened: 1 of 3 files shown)


kevinzakka:
Thanks for the PR @ZzzzzzS! Your diagnosis is correct but your test is vacuously true (with `num_envs=1` the buggy broadcast still succeeds) so closing it in favor of #1046. Note you've been credited in that PR :)

PR #1046, opened 2026-06-04
Fix CircularBuffer backfill broadcast for >2D data

kevinzakka:
`CircularBuffer.append` crashed with a broadcast mismatch when the buffered data had more than two dimensions. The first-push backfill hard-coded a 3-D broadcast, which misaligns the batch axis with the trailing axis once `DcMotorActuator`'s variable delay stacks targets into a 3-D tensor. It now derives the broadcast shape from `data.ndim` so backfill works for any rank. The regression came from #1031, which swapped the dimension-agnostic mask for a sync-free `torch.where` but baked in 3-D views.

Supersedes #1045 by @ZzzzzzS, who diagnosed the bug and wrote an equivalent fix (credited as co-author).

### DIFF
--- a/src/mjlab/utils/buffers/circular_buffer.py
+++ b/src/mjlab/utils/buffers/circular_buffer.py
@@ -208,10 +208,11 @@ class CircularBuffer:
     self._buffer[self._pointer] = data
 
     # Backfill entire history with first frame for newly initialized batches.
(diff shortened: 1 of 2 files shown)


rdeits-bd (approved):
Ah sorry about that and thanks for the fix. I don't see any new CUDA sync issues after this change.

PR #1049, opened 2026-06-04
`RecorderManager` uses a `dict[str, RecorderTerm]` instead of two lists; Add `RecorderManager.get_term()`

bd-adaniele:
# Summary

The `RecorderManager` class used two lists of the same length to store the `RewardTerm`s and their names. This PR replaces those lists with a single `dict[str, RewardTerm]`. It also introduces a public `get_term()` method so that we can avoid the illegal private access to `_terms`. The `get_term()` method is common in all other managers.


tests/test_recorder_manager.py
```diff
@@ -340,6 +342,45 @@ def test_null_recorder_manager_no_ops() -> None:
...
+
+
+def test_get_term_returns_registered_instance(mock_env: Mock) -> None:
+  """get_term returns the live term the manager dispatches to, not a copy."""
+  cfg = {"recorder": RecorderTermCfg(func=_CountingRecorder, params={})}
+  manager = RecorderManager(cfg, mock_env)
+  term = manager.get_term("recorder")
+  assert isinstance(term, _CountingRecorder)
+
+  manager.record_post_step()
```
bd-adaniele, line 354:
```suggestion
  assert isinstance(term, _CountingRecorder)
  manager.record_post_step()
```

kevinzakka (approved):
Thank you @bd-adaniele!

PR #1050, opened 2026-06-05
Regenerate MuJoCo type stubs and keep them in sync via CI

### LINKED ISSUE #1048: MuJoCo typing stubs are stale
Running `typings/generate_mujoco_stubs.sh` on `main` makes changes to the following files:
```
typings/mujoco/__init__.pyi
typings/mujoco/_constants.pyi
typings/mujoco/_enums.pyi
typings/mujoco/_functions.pyi
typings/mujoco/_specs.pyi
typings/mujoco/_structs.pyi
typings/mujoco/glfw/__init__.pyi
typings/mujoco/rendering/classic/gl_context.pyi
typings/mujoco/rendering/classic/renderer.pyi
```

I assume that the typing stubs are intended to be in sync with the version of `mujoco` in use. It looks like the stubs weren't updated the last time that MuJoCo was updated (commit `e1f2883`). This can be fixed by running `typings/generate_mujoco_stubs.sh`

kevinzakka:
The bundled MuJoCo type stubs had drifted from the installed mujoco version. This regenerates them and adds a CI job that regenerates and diffs them so they can't go stale again. To make that check deterministic across platforms, the generation script now drops the OpenGL backend stubs (which bake in mujoco.cgl on macOS vs mujoco.glfw on Linux and aren't needed to type-check mjlab) and strips machine-specific paths and handles from the output. Run make stubs to update them.

Fixes #1048


kevinzakka:
cc @bd-hberger

PR #1057, opened 2026-06-13
Add `lock_joints` spec wrapper

louislelay:
Adds a  higher-order function that wraps a `spec_fn` to delete specified joints before compilation, permanently welding them in place. This is useful for freezing degrees of freedom you don't want in the action space (e.g. disabling an actuated head). Raises if a requested joint name doesn't exist in the spec.


kevinzakka:
@louislelay There are a few bugs in this PR + we need to discuss the implementation a bit more.

PR #1060, opened 2026-06-18
Add `reduce="max"` to `MetricsTermCfg`

louislelay:
`MetricsTermCfg.reduce` only supported `"mean"` and `"last"`. This adds `"max"`, which reports the episode-peak value — useful for peak power, peak contact force, etc. Without it, users need stateful wrapper classes that maintain a running `torch.maximum` and abuse `reduce="last"`. With `reduce="max"`, a plain function suffices. Works with both step and `per_substep` terms (tracks the max of the step-averaged value). The `_episode_max` buffer is only allocated for terms that use `"max"`.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -8,6 +8,8 @@ Upcoming version (not yet released)
 Added
 ^^^^^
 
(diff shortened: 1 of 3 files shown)


kevinzakka:
@claude

kevinzakka (approved):
Thanks @louislelay!

PR #1061, opened 2026-06-21
Optimize mjlab cache hot paths

cmjang:
## Summary

This PR improves several mjlab training hot paths by reducing redundant tensor work and reusing data that is already valid within the same simulation generation or sensor update.

The changes focus on cache-safe performance improvements:

- Add generation-based memoization for derived `EntityData` quantities that are recomputed frequently from forward-derived simulation fields.
- Invalidate the entity data cache through the simulation data generation counter after `step()`, `forward()`, `reset()`, and model constant recomputation.
- Avoid unconditional observation term clones before concatenation, while preserving clone protection for in-place `clip`/`scale` processing and non-concatenated raw outputs.
- Reuse contact sensor data extracted during `update()` for air-time/history bookkeeping instead of extracting the same contact tensors again on `sensor.data` access.
- Cache fixed delay/history buffer tensors to avoid repeated allocations in common observation delay/history paths.

## Performance

Benchmarks were run on CUDA with an NVIDIA RTX 3090 after rebasing onto the latest `main`.

- End-to-end Go1 flat velocity environment, isolating `EntityData` cache impact:
  - Before: `90.2k env-steps/s`
  - After: `98.0k env-steps/s`
  - Speedup: `1.087x` (`~8.7%`)
- Observation clone+concat hot-path microbenchmark:
  - Before: `0.3715s`
  - After: `0.0972s`
  - Speedup: `3.82x`
- Repeated derived `EntityData` reads microbenchmark:
  - Before: `1.5719s`
  - After: `0.0061s`
  - Speedup: `256.5x`
- Contact sensor extract/reuse microbenchmark:
  - Before: `0.3416s`
  - After: `0.1770s`
  - Speedup: `1.93x`

The microbenchmarks measure individual hot paths and should not be interpreted as whole-training speedups. The end-to-end Go1 number is the most representative full environment measurement from this PR.

## Correctness and Safety

The cache is only used for derived quantities that depend on forward-derived simulation fields such as `xpos`, `xquat`, `xipos`, `cvel`, `subtree_com`, and geom/site pose fields. Direct qpos/qvel-like passthrough fields are intentionally not cached.

Additional tests cover cache invalidation after `forward()` and ensure observation scaling does not write through into source tensors when an observation term returns a view.

## Validation

- `uv run pytest -q` (`957 passed`)
- `uv run ty check`
- `uv run pyright`
- `uv run ruff format --check ...`
- `uv run ruff check ...`
- `git diff --check origin/main..HEAD`

### DIFF
--- a/src/mjlab/entity/data.py
+++ b/src/mjlab/entity/data.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
-from typing import TYPE_CHECKING, Sequence
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Callable, Sequence
 
 import mujoco_warp as mjwarp
 import torch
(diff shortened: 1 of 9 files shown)


kevinzakka:
Hi @cmjang, thanks for your contribution. I’ve experimented with similar optimizations over the past few months, and my conclusion was that the performance gains did not justify the added complexity and extra code. Additionally, I think these changes tend to make the code harder to read and maintain. Given that tradeoff, I don’t think this PR is worth merging as-is. I appreciate you taking the time to explore this though :)

PR #1081, opened 2026-07-01
Expose MuJoCo Warp broadphase settings in MujocoCfg

bd-pdomanico:
Mujoco warp has broadphase algorithm options that do not exist in `mjOption` as documented here: https://mujoco.readthedocs.io/en/latest/mjwarp/#options

These are set based on some heuristic thresholds in `mujoco_warp` here: https://github.com/google-deepmind/mujoco_warp/blob/af9d54ad63476e1f88118e5971f08835c3b15b62/mujoco_warp/_src/io.py#L601-L606

This PR exposes the settings in mjlab so that they can be tuned for per-task performance. These options default to `None` in which case we fallback to the `mujoco_warp` heuristics

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -11,6 +11,9 @@ Added
 - Added ``MeshCfg``, a spec editor that matches mesh assets by name and edits
   their asset-level attributes. The first attribute is ``maxhullvert``, which
   caps the collision convex hull's vertex count to lower narrowphase cost.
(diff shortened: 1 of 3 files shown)


kevinzakka (approved):
Hey thanks for this PR!

I made a small style change by moving `broadphase` and `broadphase_filter` to SimulationCfg since they are mjwarp-only settings with no equivalent in `mujoco.MjOption`.

bd-pdomanico:
Looks great. Thanks, @kevinzakka !

PR #1087, opened 2026-07-07
Add material id randomization 

bd-pmorais:
This PR adds material randomization. It's implemented as a thin wrapper around a new `_randomize_categorical_field` similar to `_randomize_model_field` and `_randomize_quat_field` but for uniform sampling of integer fields. I originally also implemented mat_texid randomization this way, but it turns out mujoco's renderer bakes the texture ID mapping into the render context and never rereads mat_texid live (even though the warp renderer _does_ reread it correctly). Since it's not supported by MujocoViewer currently I decided to leave it out to avoid inconsistent behavior between the viewer and the warp renderer.


kevinzakka (approved):
Thanks @bd-pmorais! I strengthened the tests, otherwise LGTM!

PR #1088, opened 2026-07-07
Enable skybox rendering in camera sensors

bd-pmorais:
mujoco_warp recently [added](https://github.com/google-deepmind/mujoco_warp/pull/1308) skybox rendering support, this PR just sets that flag to true during sensor context creation


kevinzakka (approved):
Thanks again!

PR #1090, opened 2026-07-09
Add support for texid randomization

bd-pmorais:
Similar in principle to #1087. Left as a draft until if/when [#3387](https://github.com/google-deepmind/mujoco/pull/3387) is accepted into mujoco as without that change texture randomization wont properly show up in the mujoco viewer.


bd-pmorais:
Superseded by #1111

PR #1091, opened 2026-07-14
Fix TorchArray world-dim expansion for mujoco_warp 3.10.0.2

qiayuanl:
mujoco_warp 3.10.0.2 (released July 13) allocates world-shared model fields as real size-1 arrays instead of stride-0 broadcast views. TorchArray detected shared fields via `stride(0) == 0`, so it stopped expanding them to nworld, and any multi-env indexing of fields like `soft_joint_pos_limits` raised `IndexError: index 1 is out of bounds for dimension 0 with size 1` during resets. This currently breaks the latest-deps CI job (7 failures in test_auto_reset.py and test_entity_data.py) on any push, and the next nightly will hit it too.

The fix accepts mjwarp's `_is_batched` marker in addition to the stride-0 check when deciding to expand a size-1 leading dim. The marker is set on all world-batched Model/Data arrays since mujoco_warp PR #1465; the stride check is kept for older versions that don't mark all arrays. Non-batched fields whose leading dim happens to be 1 (e.g. `jnt_type` of a single-joint model) carry neither signal and are still left alone.

Verified locally: with mujoco-warp 3.10.0.2 the previously failing suites pass (47 passed), and the full fast suite passes on locked 3.10.0.1 (976 passed).

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -27,6 +27,10 @@ Changed
 Fixed
 ^^^^^
 
+- Fixed ``TorchArray`` not expanding world-shared model fields to ``nworld``
+  with mujoco_warp 3.10.0.2, which allocates them as real size-1 arrays
+  instead of stride-0 broadcast views. Multi-env indexing of fields like
+  ``soft_joint_pos_limits`` raised ``IndexError`` during resets.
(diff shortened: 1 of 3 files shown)


kevinzakka (approved):
Thanks @qiayuanl!

kevinzakka (approved):
N/A

PR #1092, opened 2026-07-14
Preserve custom geom names in TerrainGenerator

bd-pdomanico:
TerrainGenerator unconditionally renamed every geom in the terrain body to `terrain_{i}`, overwriting any name a sub-terrain function had already set.

This changes that behavior to only apply to unnamed geoms and enables mjlab events to reference and reconfigure geometries by regex matching against their names.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -247,6 +247,9 @@ Changed
 Fixed
 ^^^^^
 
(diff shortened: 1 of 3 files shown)


bd-pdomanico:
Test failure looks related to unlocked dependency updates, not this PR

kevinzakka (approved):

PR #1098, opened 2026-07-15
Expand actuator acceleration for recomputation

martinorrje:
MuJoCo Warp recomputes actuator_acc0 per world during set_const, but MjLab did not expand that field before per-environment domain randomization. Large vectorized runs could therefore write beyond the single-world allocation and fail with CUDA illegal memory accesses. This change expands actuator_acc0 with the other set_const-derived fields, documents the behavior, and adds regression coverage for the registered fields. Validated with the event-manager tests, make check, make test, and a 4,096-environment G1 training iteration with COM randomization enabled.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -14,6 +14,10 @@ Changed
 Fixed
 ^^^^^
 
+- Fixed CUDA illegal memory accesses when domain randomization triggers
+  ``set_const`` with multiple environments. ``actuator_acc0`` is now expanded
+  per environment before MuJoCo Warp recomputes it.
+
 Version 1.5.1 (July 15, 2026)
 -----------------------------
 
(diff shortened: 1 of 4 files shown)


kevinzakka (approved):

kevinzakka:
Thanks for catching this!

saikishor:
@kevinzakka as of now, the 1.5.1 seems to be broken. With this change, it is working again. Can we get another release, or should we fall back to 1.5.0 right now?

kevinzakka:
@saikishor Released 1.5.2. Could you elaborate on 1.5.1 being broken? It's only broken if you're accessing actuator_acc0 right?

saikishor:
> @saikishor Released 1.5.2. Could you elaborate on 1.5.1 being broken? It's only broken if you're accessing actuator_acc0 right?

Thank you so much. Even the Unitree G1 Velocity task training fails without this change, the same issue with our velocity training as well. Some rewards are using this change. I haven't dug into the problem because the recent commit already solved it.

PR #1099, opened 2026-07-16
Fix reflectance not getting passed to add_material

bd-pmorais:
This PR fixes a bug where the reflectance field in `MaterialCfg` wasn't getting passed to add_material and was therefore silently ignored. Also retroactively adds attribution to my other PRs :)


kevinzakka (approved):
Thanks @bd-pmorais!

PR #1107, opened 2026-07-20
Fix ONNX upload with current W&B logger

wuphilipp:
RSL-RL normalizes the deprecated `wandb` logger alias to `WandbLogWriter`, but the velocity and manipulation runners only recognized the old name. As a result, ONNX export still ran locally while upload and W&B run metadata were silently skipped. This accepts both logger identifiers, matching the tracking runner, and adds regression coverage for both task runners and both identifiers.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -19,6 +19,8 @@ Changed
 Fixed
 ^^^^^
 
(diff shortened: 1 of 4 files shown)


kevinzakka (approved):
Thanks Phil!

PR #1109, opened 2026-07-21
Add the Booster T1 robot

Prababameister:


kevinzakka:
We are not adding more robots to mjlab to keep it lean.

PR #1111, opened 2026-07-21
Add support for texid randomization

bd-pmorais:
Similar in principle to #1087. Now that [#3387](https://github.com/google-deepmind/mujoco/pull/3387) is merged into mujoco this implementation of dr.mat_texid will randomize textures consistently in the mujoco native, filament and mujoco_warp renderers.

### DIFF
--- a/docs/source/randomization.rst
+++ b/docs/source/randomization.rst
@@ -1416,12 +1416,13 @@ world-space positions directly from GPU simulation data (``cam_xpos``,
 
 .. note::
 
-   ``geom_rgba`` and ``geom_size`` DR are **not** reflected in viser. Geom
-   colors and sizes are baked into the scene's GLB meshes at construction
-   time. The underlying viser API (``add_batched_meshes_simple``) supports
-   per-instance color updates via ``batched_colors``, but this requires
-   routing color-only geoms through a different handle type than the current
-   ``add_batched_meshes_trimesh`` path. Deferred for a future update.
+   ``geom_rgba``, ``geom_size`` and ``mat_texid`` DR are **not** reflected
+   in viser. Geom colors, sizes and textures are baked into the scene's
+   GLB meshes at construction time. The underlying viser API
+   (``add_batched_meshes_simple``) supports per-instance color updates via
+   ``batched_colors``, but this requires routing color-only geoms through
+   a different handle type than the current ``add_batched_meshes_trimesh``
+   path. Deferred for a future update.
 
 
 Migrating from Isaac Lab
--- a/docs/source/randomization.rst
+++ b/docs/source/randomization.rst
@@ -274,6 +274,12 @@ and ``dr.body_ipos`` are the same function).
      - ``mat_texrepeat``
      - Texture repeat in the S/T directions
      - Only affects textured materials; values should stay positive
+   * - ``dr.mat_texid``
+     - ``mat_texid``
+     - Texture assigned to a material's ``mjtTextureRole`` slot (RGB by
+       default)
+     - Samples uniformly from ``asset_cfg.texture_names``. Use ``role`` to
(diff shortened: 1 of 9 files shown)


bd-pmorais:
@kevinzakka could you check this one out too?

kevinzakka:
Sweet, was able to run it and set it up on the YAM task.


https://github.com/user-attachments/assets/f91c3f7f-4ca8-4261-8128-05abc7c31118

kevinzakka (approved):

kevinzakka (changes requested):
PR is good but there are a few nits to address:

- test_mat_texid_partial_env_ids is effectively vacuous
- Pool-restriction is untested

tests/test_domain_randomization.py
```diff
@@ -1973,6 +1973,128 @@ def test_geom_matid_shared_random(matid_env):
...
+
+  dr.mat_texid(env, env_ids=None, asset_cfg=asset_cfg)
+
+  assigned = env.sim.model.mat_texid[:, mat_ids, role]
+  assert torch.all(torch.isin(assigned, tex_ids))
+  assert not torch.equal(assigned, original)
+  assert torch.isin(assigned, torch.unique(original), invert=True).any()
+
+
+def test_mat_texid_partial_env_ids(texid_env):
```
kevinzakka, line 2061: This does not actually test env_ids isolation. The texid_env fixture is module-scoped and each test re-seeds to 42, so the original snapshot taken after test_mat_texid_draws_from_pool already ran equals what a full re-randomization produces. If env_ids is ignored and all envs are randomized, the test still passes. It fails only when run in isolation.

Fix: make the fixture function-scoped (or snapshot from a fresh env), and also assert that the targeted envs (0, 2) changed, not just that 1 and 3 stayed fixed.

tests/test_domain_randomization.py
```diff
@@ -1973,6 +1973,128 @@ def test_geom_matid_shared_random(matid_env):
...
+      <freejoint name="free_joint"/>
+      <geom name="geom1" type="box" size="0.1 0.1 0.1" mass="1.0" material="mat_a"/>
+      <geom name="geom2" type="sphere" size="0.05" mass="0.3" material="mat_b"/>
+    </body>
+  </worldbody>
+</mujoco>
+"""
+
+
+def _make_texid_env(device, num_envs=NUM_ENVS):
```
kevinzakka, line 1999: Pool restriction is untested. Every test uses texture_names=(".*"), so torch.isin(assigned, tex_ids) is always true. A bug that sampled from all textures instead of the selected subset would pass. Add one case with a strict subset (e.g. two of three textures) and assert the excluded texture id never appears.

src/mjlab/entity/entity.py
```diff
@@ -458,6 +460,10 @@ def light_names(self) -> tuple[str, ...]:
   def material_names(self) -> tuple[str, ...]:
     return tuple(m.name.split("/")[-1] for m in self.spec.materials)
 
+  @property
```
kevinzakka, line 463: texture_names and texture_ids in SceneEntityCfg have no docstrings, unlike the sibling *_names / *_ids fields.

docs/source/randomization.rst
```diff
@@ -845,11 +851,6 @@ per-environment values.
    * - Category
      - Field(s)
      - Notes
-   * - Texture-role swapping
```
kevinzakka, line 848: mat_texid only takes effect in the native and offscreen MuJoCo renderers. The viser viewer renders flat material/geom colors and ignores mat_texid, so texture randomization is not visible there. Worth a one-line note so this is not mistaken for a bug.

kevinzakka (approved):

PR #1118, opened 2026-07-27
Add missing LightCfg fields and domain randomization

bd-pmorais:
This PR adds some fields to LightCfg that were missing but supported by both mujoco and mjwarp and adds basic domain randomization functions for them.


kevinzakka (approved):
Thanks for all the visual DR-related PRs!

PR #1119, opened 2026-07-28
Add dr.light_cutoff, dr.light_exponent and add light fields to model_sync

bd-pmorais:
Extends #1118 by adding the final two missing randomizations (`dr.light_cutoff` and `dr.light_exponent`) and also adds all the randomizable fields to `model_sync` so they show up in the native viewer.


kevinzakka (approved):
Thank you!

PR #1127, opened 2026-08-03
Add reduce option `sum` for metrics

bd-mlutter:
## Changelog
* Add `sum` as additional reduce option for `metrics`. This is beneficial for metrics that accumulate over time, e.g., traveled distance or energy.

* Raise `ValueError` if `reduce` is not in the `REDUCE_OPTIONS`. Raising the error avoids silently using `mean` when `reduce` contains a typo or similar. 

## Testing
- [x]  `make format`
- [x] `make test`

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -123,6 +123,9 @@ Version 1.5.1 (July 15, 2026)
 Added
 ^^^^^
 
+- Added ``reduce="sum"`` to ``MetricsTermCfg`` for reporting the accumulated
+  episode total (e.g. episodic reward, total distance traveled) instead of a
+  per-step average.
 - Added ``MeshCfg``, a spec editor that matches mesh assets by name and edits
   their asset-level attributes. The first attribute is ``maxhullvert``, which
   caps the collision convex hull's vertex count to lower narrowphase cost.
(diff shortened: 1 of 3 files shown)


src/mjlab/managers/metrics_manager.py
```diff
@@ -118,18 +133,34 @@ def reset(
...
         extras["Episode_Metrics/" + key] = torch.mean(
           self._episode_sums[key][env_ids] / safe_counts
         )
+
+      else:
+        msg = (
+          f"The reduce method '{reduce}' for metric '{key}' is unknown. "
+          f"Valid options are {REDUCE_OPTIONS}."
+        )
+        raise ValueError(msg)
```
kevinzakka: This is unreachable now that `__init__` validates against `REDUCE_OPTIONS`. Please delete.

docs/source/changelog.rst
```diff
@@ -123,6 +123,9 @@ Version 1.5.1 (July 15, 2026)
 Added
 ^^^^^
 
+- Added ``reduce="sum"`` to ``MetricsTermCfg`` for reporting the accumulated
```
kevinzakka: Should be under a new "Upcoming version (not yet released)" heading please.

kevinzakka (changes requested):
Thanks for the contribution, @bd-mlutter! Left 2 comments.

docs/source/changelog.rst
bd-mlutter: moved to upcoming version

src/mjlab/managers/metrics_manager.py
bd-mlutter: Partially, yes. I still do like it just for readability. I removed it and kept the `ValueError` in the `__init__`, which triggers earlier which is nicer for experiments.

kevinzakka (approved):

PR #1128, opened 2026-08-04
Add fields for loading textures from files

bd-pmorais:
This PR extends `TextureCfg` to all the fields read by the mujoco_warp renderer which allows the creation of textures from a file.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -24,6 +24,11 @@ Added
 - Added ``diffuse``, ``specular``, ``ambient``, ``active``, and
   ``attenuation`` fields to ``LightCfg`` for configuring light color and
   falloff. Contribution by @bd-pmorais.
+- Added ``random``, ``file``, ``cubefiles``, ``gridsize``, ``gridlayout``,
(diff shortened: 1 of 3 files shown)


kevinzakka (changes requested):
Thanks, this looks good overall. Two small nits in `validate` before I merge:

1. The skybox check fires even when builtin is set. MuJoCo gives builtin precedence over file (a builtin="gradient" skybox with a file still compiles to six faces), so a config like that would compile fine but get rejected here. Can you add `self.builtin == "none"` to the condition so the error only fires when the file is actually what gets rendered?
2. Now that builtin defaults to "none", something like TextureCfg(name="t", type="2d") is constructible with no builtin, no file, and no cubefiles. It passes validation but fails at compile time with MuJoCo's confusing "Cannot load 2D texture from separate files" error. Could you add a validate check that at least one texture source is set, e.g.:

```python
if self.builtin == "none" and self.file is None and self.cubefiles is None:
  raise ValueError(
    "Texture must specify a builtin pattern, a file, or cubefiles."
  )
```

kevinzakka (approved):

PR #1130, opened 2026-08-04
Bump mujoco and mujoco-warp to 3.11

bd-pmorais:
Both come from PyPI, so this is just a version bump in pyproject.toml plus a regenerated lockfile and refreshed MuJoCo type stubs.

The only source change needed is that MjsBody.explicitinertial is now typed as bool rather than int upstream, so the int() coercion in _detect_inertial_mode and the `= 1` assignments in the variants tests are updated to match.


kevinzakka (approved):
Perfect PR, thank you!

PR #1131, opened 2026-08-04
clean up `OffscreenRenderer` and `ViewerConfig`

bd-mlutter:
## Summary
The main changes are:
* Improve readability of the `OffscreenRenderer`
* add more comments to the `ViewerConfig`
* add a few minor features

## New Features:
* `viewer/offscreen_renderer.py` & `viewer/viewer_config.py`
   * add options to enable and disable `site` and `geom` groups
   * `world` camera selects the `envs` that are closest to `lookat` instead of the specified `env id`

## Open questions:
- Should the debug visualization be drawn for all selected environments or only for the first environment? Should one be able to switch between both cases using the `ViewerConfig`?

### `viewer/offscreen_renderer.py`
- Instead of mutating a shared `MjModel` the model is copied and mutated.
   - The renderer works on a `copy.copy` of the model
   - Therefore, the resetting of the model in `close()` is not necessary anymore.

- The rendered environment ids are resolved in the constructor.

- convert some methods to static methods with explicit args and return values (e.g., `setup_camera`, `computer_render_extend` and `get_environment_ids`).

- moved inferring the camera body id for the camera into a separate function, as this logic made `setup_camera` more convoluted than necessary. This function is in `viewer_config.py`

- `max_extra_envs > n_world` raises a `ValueError` instead of being silently clamped.
 
- added function `sync_data_fields` that sync between `mjWarp.Data` and `mujoco.MjData` similar to `sync_model_fields`

- `cfg.fovy` now applies to every origin type.
    - Previously ignored for the tracking cameras (`ASSET_ROOT`/`ASSET_BODY`).
    - silently ignoring the `fov` seems counterintuitive as one can just set `fov = None` and use the default

### `viewer/viewer_config.py`
- Added `geom_group` / `site_group` fields.
- Added `get_camera_body_id` function that infers the body `id` for the `ASSET_ROOT` and `ASSET_BODY` origin

- Reordered variables into semantic groups
  - First, variables that apply to all origins types
  - Second variables that are specific to different origin types

### `viewer/native/visualizer.py`
 - renamed attributes to make types more clear, i.e., add `mj_` to `mujoco` variables
 - renamed `mj_model` to `ref_mj_model` to highlight that this model is shared and not owned
 - added properties that have the old names to not break the previous public variables
 - grouped variables that are only used by the ghost visualization
 - added comments to highlight limitations and assumptions.

## Testing
- [x] Formatting (`make format`)
- [x] Unit tests (`make test`)

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -35,11 +35,31 @@ Added
 - Added ``reduce="sum"`` to ``MetricsTermCfg`` for reporting the accumulated
   episode total (e.g. episodic reward, total distance traveled) instead of a
   per-step average. Contribution by @bd-mlutter
+- Added ``ViewerConfig.geom_group`` and ``ViewerConfig.site_group`` to control
+  which MuJoCo geom and site visualization groups the offscreen renderer draws.
+  Both default to ``(1, 1, 1, 0, 0, 0)``, matching the previous behavior.
+  Contribution by @bd-mlutter
 
 Changed
 ^^^^^^^
 
 - Bumped ``rsl-rl-lib`` from 5.4.0 to 5.4.2.
+- ``ViewerConfig`` is now keyword-only. Its fields were reordered into semantic
+  groups (general options, camera placement, then origin-type-specific fields),
+  so positional construction is no longer supported.
+- ``ViewerConfig.fovy`` now applies to every ``origin_type``. It was previously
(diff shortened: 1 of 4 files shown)


kevinzakka:
Thanks Michael! This is now superseded by #1147.

PR #1139, opened 2026-08-06
Stop partial resets from advancing stateful commands in other envs

### LINKED ISSUE #1138: Stateful commands (e.g. MotionCommand) are unintentionally advanced on `env.reset`
https://github.com/mujocolab/mjlab/pull/762 added a `command_manager.compute(dt=0)` call on `env.reset`. That ends up here:

https://github.com/mujocolab/mjlab/blob/fe1e8ab625aa730a3bbafef395ce9f3ad215b08d/src/mjlab/managers/command_manager.py#L249-L251

and thus here:

https://github.com/mujocolab/mjlab/blob/fe1e8ab625aa730a3bbafef395ce9f3ad215b08d/src/mjlab/managers/command_manager.py#L109-L115

which results in `_update_command` being called for every command term (implicitly for *every* environment). 

This leads to a pretty weird behavior: If your environment has a `MotionCommand` or any other command which does something like this: 

https://github.com/mujocolab/mjlab/blob/fe1e8ab625aa730a3bbafef395ce9f3ad215b08d/src/mjlab/tasks/tracking/mdp/commands.py#L407-L408

then doing `env.reset(env_ids=env_ids)` will advance the reference motion by one tick for *all* environments, regardless of which environments you reset. This makes `env.reset` kind of unsafe to use with anything other than `env_ids=<all envs>`, since the non-reset events will have their command state corrupted. 

This makes the `auto_reset=False` workflow hard to use correctly -- the docstring suggests users call `env.reset(env_ids=...)` themselves, but that's not really a safe thing to do if you have command terms like `MotionCommand`. 

We ran into this with a loop that looked roughly like:

```py
env.cfg.auto_reset = False
while True:
  ... = env.step()
  <inspect the terminated envs>
  env.reset(env_ids=<terminated envs>)
```

which resulted in our `MotionCommand` advancing exactly twice as fast as usual! Once during `step` and then again for all envs during `reset`.  

One interpretation is that this is a bug in `MotionCommand` (and probably lots of other command terms). But it's hard to fix this cleanly with the command API because `_update_command` doesn't get an `env_ids` or a `dt` call. And we presumably *do* want to call `_update_command` for the envs which were reset, since that's the expected behavior of a command term.

kevinzakka:
`env.reset(env_ids=...)` runs `command_manager.compute(dt=0)` (added in #762 to refresh derived command state after a reset), which called `_update_command` for every environment. Stateful commands advanced on each partial reset: with `auto_reset=False`, a `MotionCommand` reference motion played at 2x speed, and non-reset robots could even be teleported by the wraparound resample.

`_update_command` now takes `env_ids`: `None` on the regular per-step update, the reset ids when called from `reset()`. The step path is unchanged; a reset only ticks the envs being reset, matching what the auto-reset path already does. The adaptive-sampling EMA folds only on the per-step update so manual resets no longer decay it faster. Command terms with the old zero-arg signature fail fast at construction with a migration hint.

Fixes #1138

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -12,6 +12,12 @@ Upcoming version (not yet released)
      and ``priority`` to be explicit instead of silently defaulting to
      MuJoCo's values, and dict values for these fields must cover every
      matched geom (add a catch-all ``".*"`` entry).
+   - ``CommandTerm._update_command`` now takes an ``env_ids`` argument:
+     ``None`` on the regular per-step update and the reset environment ids
+     when called from ``reset()``. Custom command terms must add the
+     parameter (construction raises a ``TypeError`` with migration
+     instructions otherwise) and scope any per-step state advance, such as
+     a motion frame index, to ``env_ids``.
 
 .. admonition:: Highlights
    :class: note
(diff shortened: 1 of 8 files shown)


rdeits-bd:
Thanks!! This looks good -- I just want to check in today with the other BD people who ran into this.

rdeits-bd (approved):
thanks!

PR #1141, opened 2026-08-06
Stop partial resets from advancing other envs' observation buffers

kevinzakka:
`env.reset(env_ids=...)` recomputed observations with `update_history=True`, appending a frame to every env's history buffer and advancing every delay buffer, not just the reset envs. In `auto_reset=False` loops, surviving envs got a duplicate frame per manual reset: history windows shrank and delay schedules drifted.

The buffers now expose `backfill` (fill a reset row's history with its post-reset frame without advancing global time) and `peek` (read the delayed obs without stepping the lag schedule); `reset()` routes the reset envs through them. The per-step path is unchanged. Regression tests at the buffer and env level fail on the previous code.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -64,6 +64,11 @@ Changed
 Fixed
 ^^^^^
 
+- ``reset(env_ids=...)`` no longer appends a frame to every env's observation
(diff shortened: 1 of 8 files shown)


rdeits-bd (approved):
LGTM, thank you!

PR #1163, opened 2026-08-27
Upgrade rsl-rl from 5.4.2 to 5.5.0

bd-hberger:
This PR updates `mjlab` to use [version 5.5.0](https://github.com/leggedrobotics/rsl_rl/releases/tag/v5.5.0) of `rsl-rl`, which was released today. 

Among the changes is a fix so that empirical normalization is synchronized across all GPUs when doing multi-GPU training.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -5,6 +5,11 @@ Changelog
 Upcoming version (not yet released)
 -----------------------------------
 
+Changed
+^^^^^^^
+
+- Bumped ``rsl-rl-lib`` from 5.4.2 to 5.5.0. 
+
 Version 1.6.0 (August 8, 2026)
 ------------------------------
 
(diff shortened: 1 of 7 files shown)


kevinzakka (approved):

bd-hberger:
The `rsl-rl` update included https://github.com/leggedrobotics/rsl_rl/pull/230, which changed how you check whether the logger is using wandb. The attribute `logger.logger_type` no longer exists, and you need to check the type of `logger.writer` instead. I made this change and also noted it in the changelog.

kevinzakka (approved):
Thanks @bd-hberger.

bd-hberger:
> The `rsl-rl` update included [leggedrobotics/rsl_rl#230](https://github.com/leggedrobotics/rsl_rl/pull/230), which changed how you check whether the logger is using wandb. The attribute `logger.logger_type` no longer exists, and you need to check the type of `logger.writer` instead. I made this change and also noted it in the changelog.

By the way, this change was only caught by the type check, not by the pytests, because the pytests always mock out the logger. This seems like a potential gap in test coverage

PR #1167, opened 2026-08-28
Cap wandb below 0.29 to fix crash on startup with --logger wandb

k1000dai:
## Problem

wandb 0.29.0 (released 2026-08-26) removed the deprecated `start_method` setting
([changelog](https://github.com/wandb/wandb/blob/main/CHANGELOG.md), wandb/wandb#12581).
`wandb.Settings` is a strict pydantic model, so passing the removed argument now raises
instead of being ignored.

rsl-rl-lib (including the currently pinned 5.5.0) still passes it in
`rsl_rl/utils/wandb_log_writer.py`:

```python
settings=wandb.Settings(start_method="thread"),
```

Since `pyproject.toml` has `wandb>=0.22.3` with no upper bound, a fresh install now
resolves wandb 0.29.0, and any training run with `--logger wandb` crashes at startup:

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -12,6 +12,12 @@ Changed
   attribute of the ``rsl_rl.utils.Logger``, so code that previously checked
   ``logger.logger_type`` must instead check the type of ``logger.writer``.
 
+Fixed
+^^^^^
+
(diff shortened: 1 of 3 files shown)


kevinzakka (approved):
Thanks!

PR #1172, opened 2026-08-31
Fix BuiltinPdActuator ctrl layout when sort_actuators=True

### LINKED ISSUE #1170: BuiltinPdActuator applies swapped pos/vel targets when EntityCfg.sort_actuators=True
When an entity uses `EntityCfg(sort_actuators=True)`, `Entity._add_actuators` flattens each actuator config to per-target calls: `edit_spec(spec, [name])` once per target. `BuiltinPdActuator.edit_spec` appends one `<position>` and one `<velocity>` element per call, so the elements end up interleaved in the compiled model:

```
sort_actuators=False: [joint1_pd_pos, joint2_pd_pos, joint1_pd_vel, joint2_pd_vel]
sort_actuators=True:  [joint1_pd_pos, joint1_pd_vel, joint2_pd_pos, joint2_pd_vel]
```

But `compute()` still returns `torch.cat((position_target, velocity_target))` (the `[pos..., vel...]` block layout), and `ctrl_ids` are derived from the appended order, so position targets are written to velocity actuators and vice versa. `dr.pd_gains` slices `ctrl_ids[:n]` / `ctrl_ids[n:]` under the same block assumption and randomizes the wrong elements too.

Repro (2-joint robot, kp=100, kd=10, pos target [0.1, 0.2], vel target [1.0, 2.0], joints at rest):

```
sort_actuators=False: qfrc_actuator = [20.0, 40.0]   # = kp*pos + kd*vel, correct
sort_actuators=True:  qfrc_actuator = [12.0, 120.0]  # scrambled
```

No test covers the combination: `sort_actuators` is only exercised with single-ctrl actuators, and the BuiltinPd layout test uses the default unsorted path.

hadelan:
With sort_actuators=True, Entity._add_actuators calls edit_spec once per target, so the `<position>` and `<velocity>`elements of BuiltinPdActuator end up interleaved in the spec as [pos_0, vel_0, pos_1, vel_1]. compute() still emits [pos..., vel...] and dr.pd_gains slices ctrl ids under the same block assumption, so position targets were applied to velocity actuators and vice versa, silently corrupting the PD torques.

The actuator now records the position and velocity elements separately and restores the block layout in initialize() before the base class derives ctrl ids from it.

Fixes #1170.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -15,6 +15,11 @@ Changed
 Fixed
 ^^^^^
 
+- ``BuiltinPdActuator`` now keeps its ``[pos..., vel...]`` ctrl layout when the
+  entity uses ``sort_actuators=True``. The per-target ``edit_spec`` calls
+  interleaved the position and velocity elements, so position targets were
+  written to velocity actuators and vice versa, silently corrupting the PD
+  torques (and the ``dr.pd_gains`` randomization).
 - Capped ``wandb`` below 0.29, which removed the ``start_method`` setting still passed
   by ``rsl-rl-lib`` and crashed training runs launched with ``--logger wandb``.
 - ``distribution="gaussian"`` domain randomization now draws an independent value per
--- a/src/mjlab/actuator/builtin_actuator.py
+++ b/src/mjlab/actuator/builtin_actuator.py
@@ -11,6 +11,7 @@ from enum import IntEnum
 from typing import TYPE_CHECKING
 
 import mujoco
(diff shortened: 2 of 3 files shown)


hadelan:
Hi @kevinzakka , a friendly ping on this PR — CI is green and the branch is up to date with main. Happy to make any changes if needed. Thanks!

hadelan:
Hi @kevinzakka , just wanted to follow up on this PR since it’s been about several weeks. Whenever you have some time, I’d really appreciate a review. Thanks!

PR #1173, opened 2026-08-31
Fix flat patch sampling collapsing to the center when patch_radius=0

### LINKED ISSUE #1171: FlatPatchSamplingCfg(patch_radius=0) collapses all patches to the sub-terrain center
`find_flat_patches_from_heightfield` excludes border pixels with:

```python
radius_pixels = int(np.ceil(cfg.patch_radius / horizontal_scale))
...
valid_mask[-radius_pixels:, :] = False
valid_mask[:, -radius_pixels:] = False
```

With `patch_radius=0`, `radius_pixels == 0` and numpy's `arr[-0:]` is `arr[0:]`, so the two negative-index slices clear the **entire** valid mask instead of nothing. Every pixel counts as invalid, the function hits the no-valid-patches fallback, and all `num_patches` samples come back as the sub-terrain center.

Repro: all-flat 80x80 heightfield with `FlatPatchSamplingCfg(num_patches=8, patch_radius=0.0)` returns 8 identical patches at the center; `patch_radius=0.5` returns 8 distinct patches. Anything spawning from flat patches (`reset_root_state_from_flat_patches`) then places every env on the same point, with no error or warning.

`patch_radius=0` is a legitimate configuration meaning point-wise flatness samples (a single-pixel footprint always lies within bounds), and nothing validates against it.

hadelan:
The edge-exclusion mask in find_flat_patches_from_heightfield used negative-zero slicing: valid_mask[-0:] is valid_mask[0:], so with radius_pixels == 0 the whole valid mask was cleared even though a zero-radius footprint is valid on every flat pixel. Every patch then fell back to the sub-terrain center.

The exclusion is now skipped when radius_pixels == 0.

Fixes #1171.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -15,15 +15,14 @@ Changed
 Fixed
 ^^^^^
 
-- ``FlatPatchSamplingCfg(patch_radius=0)`` no longer collapses every patch to
-  the sub-terrain center. The edge-exclusion mask used negative-zero slicing
-  (``arr[-0:]`` is ``arr[0:]``), which cleared the whole valid mask and left
-  the center fallback as the only sampled point.
 - Capped ``wandb`` below 0.29, which removed the ``start_method`` setting still passed
   by ``rsl-rl-lib`` and crashed training runs launched with ``--logger wandb``.
 - ``distribution="gaussian"`` domain randomization now draws an independent value per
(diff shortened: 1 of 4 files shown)


kevinzakka (approved):
Thanks for catching this. I pushed a small follow-up to drop the branch and tighten the test.

PR #1174, opened 2026-09-01
Add teacher-student distillation with DAgger and RL fine-tuning

JonasFrey96:
This adds a teacher-student distillation pipeline following the distill-then-finetune recipe of "Parkour in the Wild" (arXiv:2505.11164), built on the DAgger-style Distillation algorithm that ships with rsl-rl 5.5. MjlabDistillationRunner lets the student act in the environment (with zero-mean exploration noise from its action distribution) while frozen teachers label every visited state, and loads teacher weights directly from ordinary PPO checkpoints via a teacher_checkpoints config entry. A MultiTeacherModel distills several skill- or terrain-specific experts into one generalist student, dispatching each environment to its expert through an integer observation group such as the new terrain_type term, so the assignment travels with the observations and works both in rollouts and replayed batches. For the RL fine-tuning stage, init_checkpoint initializes a PPO actor from a distilled student while dropping the distribution parameters so a reduced initial action std applies, and CriticWarmupPPO freezes the actor for the first N updates so the freshly initialized value function cannot destroy the distilled behavior.

Two example tasks demonstrate the pipeline on G1 velocity: Mjlab-Velocity-Rough-Unitree-G1-Distill compresses the height-scan teacher into a blind recurrent (LSTM) student, and Mjlab-Velocity-Rough-Unitree-G1-Distill-Finetune fine-tunes it with asymmetric PPO on the privileged critic observations. ONNX export metadata now takes the policy observation group as a parameter, since distilled policies consume the "student" group rather than "actor"; the generic task-config tests were relaxed accordingly. One tradeoff worth noting: the fine-tune actor intentionally reloads only model and normalizer weights, never the optimizer or std, which makes resuming a distillation run and seeding a new one from an old student use the same code path. Includes tests covering the DAgger loop, multi-teacher dispatch and loading, the init-checkpoint mapping, and the critic warmup, plus a documentation page describing the three-stage workflow.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -78,6 +78,7 @@ Table of Contents
    :caption: Training & Debugging
 
    source/training/rsl_rl
+   source/training/distillation
    source/viewers
    source/training/distributed_training
    source/training/cloud
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -5,6 +5,24 @@ Changelog
 Upcoming version (not yet released)
 -----------------------------------
 
+Added
+^^^^^
+
+- Teacher-student distillation and RL fine-tuning support following
+  `arXiv:2505.11164 <https://arxiv.org/abs/2505.11164>`_.
+  ``MjlabDistillationRunner`` wraps rsl-rl's DAgger-style ``Distillation``
+  algorithm with mjlab checkpoint handling (``teacher_checkpoints`` loads
+  frozen teachers from PPO checkpoints), ``MultiTeacherModel`` distills
+  several experts into one student with per-env expert assignment,
+  ``init_checkpoint`` initializes a PPO actor from a distilled student with
+  a reduced action std, and ``CriticWarmupPPO`` freezes the actor while the
+  critic warms up. New example tasks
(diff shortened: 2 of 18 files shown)


louislelay:
Hi @JonasFrey96, thanks for your interest in mjlab! Let's confirm with @kevinzakka, but I think we won't accept the PR. OOC, I'd still love to see the results you were able to get with the mjlab robots! If you have videos, feel free to share them.

The reason is mainly that we prefer to keep the repository lean and easy to maintain with a minimal amount of robots and tasks. That said, we definitely recommend that you create an external project!

Here's how to create the project, using mjlab as a dependency: https://mujocolab.github.io/mjlab/main/source/installation.html#method-1-use-mjlab-as-a-dependency-uv

And here's the Anymal C project as an example of how to structure it: https://github.com/mujocolab/anymal_c_velocity

Once you're done, feel free to submit a PR to add it to the docs, as we keep a list of all the open-source projects using mjlab. You can also share it in the discussions!

kevinzakka:
I’ll take a look as soon as I can, if it’s useful and lean we can merge it!

Issue #934, opened 2026-04-22
[Question] Fairness of `terrain_levels_vel` curriculum progression function for low speed locomotion task


OscarMrZ:
Good morning!

I am experimenting with stairs terrain curriculums for low speed locomotion (ranging from 0.1 to 0.6 m/s), and I have realized my environments do not progress as I would have expected. I am using the STAIRS_TERRAINS_CFG terrain generator cfg, with tile size reduced to 6x6m. The problem is that while some envs do reach max difficulty, the average stagnates:

<img width="1146" height="421" alt="Image" src="https://github.com/user-attachments/assets/87d548ab-6275-401b-a9f1-a8650db682b6" />

<img width="1146" height="421" alt="Image" src="https://github.com/user-attachments/assets/15c97017-02e4-411c-9829-565390b680ab" />

As far as my understanding goes, as the policy gets better in handling the rougher terrains, more and more envs should be promoted towards the maximum difficulty. I understand that when a given env reaches the maximum it gets randomly reassigned to any row, but I think this should make the average level hover close to the max, not in the middle.

With this in mind, I was wondering if the promotion function `terrain_levels_vel` is being "fair" promoting/demoting the environments, considering the `move_up` and `move_down` conditions, and I see three possible issues that I would really appreciate some clarification about.

1. Only envs that moved an absolute distance wrt the center of the terrain tile get promoted. This creates a somewhat hidden dependency between episode time, terrain tile size and target velocity. For example, in my case, when commanding velocities in the range (-0.2, 0.2) and considering an episode lasts for 10s, the envs will never be promoted, as with perfect tracking, the robot will only travel 2m in the best of cases. So I wonder, should this be fixed just by adjusting the tile size to a more sensible value or could the `move_up` maybe be improved considered the `expected` travel distance instead?

2. In the case of also having a yaw velocity component, the problem becomes even worse, as even with higher velocities, the robot will do a circle and probably not move much wrt the origin in absolute terms. This may even lead to an unfair demotions as the `move_down` logic only considers x and y speed for calculating the distance the robot should have traveled.

3. In my case, velocity commands get resampled every 3-8 s during an episode, but the function just takes the last one (afaik). I can think of two problems because of this:

   - The robot may receive a positive command first and then a negative one, finishing close to the starting point with perfect tracking -> Unfairly not promoted and maybe even demoted.
[...]

PR #1184, opened 2026-09-16
Add history to commands

OscarMrZ:
Hello!

This PR adds history to `CommandTerm` through an auxiliary `CommandHistory` class. This is an opt-in feature disabled by default, so this PR doesn't introduce any behavior changes. All derived commands can enable the feature when defining the tasks when needed.

Afaik, this feature is needed to implement a better terrain curriculum progression function, as discussed in https://github.com/mujocolab/mjlab/issues/934, as it allows to integrate/average the velocities that were actually asked to the robot and fairly judge promotion/demotion. If it is okey by you, I'll address that in a separate MR once we settle in a command history interface.

Thanks a lot for your time!

Óscar


OscarMrZ:
Closing as #1188 implements commanded distance without history.

PR #1185, opened 2026-09-16
Update wandb & rsl-rl-lib

nim65s:
Hi,

https://github.com/leggedrobotics/rsl_rl/pull/232 is now available in 5.5.1, so https://github.com/mujocolab/mjlab/pull/1167 is no longer needed

cc @k1000dai


kevinzakka (approved):
Thanks!

PR #1187, opened 2026-09-18
Simulation backend registry and Mujoco CPU based backend prototypes

bd-hberger:
# Summary

This PR consists of two main additions:
1. A protocol for defining new `mjlab` simulation backends  and a registry to make them available in `mjlab`. Simulation backends can be defined inside or outside of `mjlab` itself.
2. `MujocoSimulation` and `MujocoSimulationWithKinematics` prototypes that implement the simulation backend protocol using CPU MuJoCo instead of MuJoCo Warp. 
    1. `MujocoSimulation` utilizes the `mujoco.rollout` threadpool, but it only provides back state and sensordata (unlike the standard MuJoCo Warp backend, which provides back a full `mjwarp.Data`). Tasks that wish to use derived kinematic quantities must use sensors.
    2. `MujocoSimulationWithKinematics` begins to expand the remaining mjdata as it has been needed. This is slower than `MujocoSimulation`.

> [!IMPORTANT]
> This PR is not intended to merge as-is.
> The rendering protocol and registry are production-ready, but the Simulation backends should be viewed as prototypes only. A simulation backend based on [mjbatch](https://github.com/kevinzakka/mjbatch) should be the successor to both of these prototypes. `mjbatch` can provide access to arbitrary fields from `mjdata`, whereas `mujoco.rollout`, which these prototypes use, only provides access to state (`mjSTATE_FULLPHYSICS`) and sensor data.

Notes:
* Ruff formatting and unit tests pass on this PR, but type-checking does not.
* Due to limitations imposed by the use of `mujoco.rollout`, most tasks defined for the `mjwarp` backend will not work with the new backends introduced in this PR. Example tasks using these backends will be provided soon in a separate PR.


bd-hberger:
Since this PR isn't intended to merge as-is, I am closing this PR and opening an identical one in our fork (https://github.com/boston-dynamics/mjlab/pull/1)

